PR #30, opened 2016-06-02
Updated some wording

johanneslumpe:
@rofrischmann I have noticed some inconsistencies with upper/lowercase regarding `Renderer`, `Keyframe` and `Selector` - I feel that they should all just be lowercase. Will change if you agree.

### DIFF
--- a/docs/Workflow.md
+++ b/docs/Workflow.md
@@ -64,7 +64,7 @@ const style = {
 
 #### 2. Validation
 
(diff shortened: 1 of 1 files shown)


rofrischmann:
I really appreciate your help! Was just about to call for feedback. Especially the documentation which has evolved and changed a lot during its creation. I will soon try to add more middleware features to provide great DX. React bindings are probably coming today :P 
Then we should share Fela as much as possible :)

johanneslumpe:
@rofrischmann Feel free to merge this in. I will go through the rest of the docs on soon. I'm so looking forward to the react bindings :D

PR #32, opened 2016-06-04
Some more doc updates

johanneslumpe:


docs/GettingStarted.md
```diff
@@ -65,16 +65,19 @@ const selector = ({ color = 'red' } = {}) => ({
 })
 
 // OR operator
-const selector = props => ({
-  color: props.color || 'red'
-})
+const selector = props => {
+  props = props || {}
```
rofrischmann: This is not necessarily required as Fela uses an empty object as default, but to be sure it does not break one should use it, I guess :P

rofrischmann:
Amazing :) The docs really get into shape now. Also feel free to request more information if you feel something is not enough explained.

docs/GettingStarted.md
johanneslumpe: Ah, well I tried to get both examples on par, as the example above uses the default parameters and also sets a default object. The fact that Fela passes in an empty object is something we should note in the docs and then remove those default objects I think.

rofrischmann: Yeah perhaps. It just simplifies rendering base selectors as you can do `fela.render(selector)` instead of `fela.render(selector, {})`. Same goes for plugins which uses an empty array by default, but I am not sure if we should not remove the plugins parameter from render at all since you can pass them to the renderer directly. I do not see any use case where it would be required to render with some additional plugins which are only required for a single selector.

PR #38, opened 2016-06-04
Bind `render` to renderer instance

johanneslumpe:
Required changes for https://github.com/rofrischmann/react-fela/pull/3


rofrischmann:
I am not sure if we should use this syntax. It is yet not part of ECMAScript's specification (afaik)

PR #42, opened 2016-06-13
Feature/this less renderer

johanneslumpe:
Removed `this` from the renderer so we do not have to bind the render* functions.

### DIFF
--- a/modules/renderer/Renderer.js
+++ b/modules/renderer/Renderer.js
@@ -114,7 +114,7 @@ export default function Renderer(config = { }) {
      * @param {FontFace} fontFace - fontFace which gets rendered
      * @return {string} fontFamily reference
      */
-    renderFont(family, files, props = { }) {
+    renderFont(family, files, properties = { }) {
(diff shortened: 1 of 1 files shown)


modules/renderer/Renderer.js
```diff
@@ -114,15 +114,15 @@ export default function Renderer(config = { }) {
      * @param {FontFace} fontFace - fontFace which gets rendered
      * @return {string} fontFamily reference
      */
-    renderFont(family, files, properties = { }) {
+    renderFont(family, files, props = { }) {
```
rofrischmann: I actually used `properties` to show that those are **not** like the other `props` but rather a set of only 5 valid properties that can be set additionally

johanneslumpe: Ah ok. I can revert that, not a problem - are you cool with the rest of the changes?

rofrischmann: Yeah just revert that, the rest seems fine :)

johanneslumpe:
@rofrischmann updated!

PR #48, opened 2016-06-22
More doc updates

johanneslumpe:


johanneslumpe:
@rofrischmann here's the next batch for now.

Issue #50, opened 2016-06-27
Reuse markup from server rendering in client


dustin-H:
Actually when rendering fela styles on the server, the new created fela renderer overwrites the styles which are already in DOM.

This causes the browser to rerender all styles which leads to a slight flash of the hole window, because the styles disappear and appear again.
This is quiet annoying.

So in my opinion a solution would be to create two methods in `renderer` which can export/import the renderer state.

Cheers
Dustin

johanneslumpe:
@dustin-H You are right! A way to hydrate the client-side renderer is already on the todo list :)

dustin-H:
Thanks @johanneslumpe, I will try to implement this tomorrow morning.

rofrischmann:
@dustin-H and I found out we actually do not have to reuse those styles for SSR as the renderer can just rehydrate and (only if styles changed) update on componentDidMount.

PR #51, opened 2016-06-28
Reuse server renderer state on the client side

dustin-H:
Solves #50 

**How it works:**

On the server-side I'm counting all `_emitChange` calls. When I have got the same styles rendered on the server and on the client, the number of `_emitChange` calls is the same. So after passing the state to the client-side renderer it ignores the first `n` `_emitChange` calls and does not apply the styles initially to the mount node. 

So the styles do not get unnecessary rerendered in the users browser, which improves the performance and user-experience.


dustin-H:
Forget about this... I will push a better solution later.

PR #52, opened 2016-06-28
Reuse server renderer state on the client side

dustin-H:
Obviously this doesn't provides any state. However it enables everybody to do this by their own.

With this change initially provided styles will not get overwritten.

# How to achieve state sharing with it?

There are two ways to reuse/recreate the state on the client-side. The first one is a little `hacky` and could lead to performance issues in the browser, however it reduces network traffic:

## 1. Counting `_emitChange` calls

On the server-side I'm counting all `_emitChange` calls. When I have got the same styles rendered on the server and on the client, the number of `_emitChange` calls is the same. So after passing the state to the client-side renderer it ignores the first `n` `_emitChange` calls and does not apply the styles initially to the mount node.

So the styles do not get unnecessary rerendered in the users browser, which improves the performance and user-experience.

To achieve this you need to overwrite the fela ´renderer._emitChange´ function like this to count all it's calls:

```js
let counter = 0
let renderer = createRenderer(config)
let emitChange = renderer._emitChange
renderer._emitChange =  () => {
   counter++
   emitChange()
}
```

Then you need to pass the counter number to the client somehow. 
For example you can insert this into your html:

```
<script>var __GLOBAL__.felaState = ${counter};</script>
```

On the client side also need to overwrite the `_emitChange` method:

```js
let offset = __GLOBAL__.fetaState
let renderer = createRenderer(config)
let emitChange = renderer._emitChange
renderer._emitChange =  () => {
   if(offset > 0){
      offset--
      return
   } 
   emitChange()
}
```

This cancels the first `n` `_emitChange` calls.

> Note: The `mountNode` has to be the same as you rendered the static css into at the server-side.

## 2. Transport the hole renderer

This causes a lot more network traffic. Basically the styles will be send 3 times. (Static styles 4 times)

From the server you just return the hole renderer somewhere like this: 

```
<script>var __GLOBAL__.felaState = ${JSON.stringify(renderer)};</script>
```

At the client-side you have to overwrite some properties of the new renderer:

```js
let renderer = createRenderer(config)
renderer.fontFaces = __GLOBAL__.felaState.fontFaces
renderer.keyframes = __GLOBAL__.felaState.keyframes
renderer.statics = __GLOBAL__.felaState.statics
renderer.rules = __GLOBAL__.felaState.rules
renderer.mediaRules = __GLOBAL__.felaState.mediaRules
renderer.rendered = __GLOBAL__.felaState.rendered
renderer.base = __GLOBAL__.felaState.base
```

No matter which way you choose this PR is necessary for both. Hope you understand and merge ;).

Cheers
Dustin

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -3,7 +3,7 @@
 Last but not least this chapter ships the API reference in detail. Despite explaining every single method its arguments and return value, you will also find some tips on some pages.
 
 ## Top-Level API
-* [`createRenderer([config])`](api/createRenderer.md)
+* [`createRenderer([config, initialState])`](api/createRenderer.md)
 * [`render(renderer, mountNode)`](api/render.md)
 * [`combineRules(...rules)`](api/combineRules.md)
 * [`enhance(...enhancers)`](api/enhance.md)
--- a/docs/API.md
+++ b/docs/API.md
@@ -3,7 +3,7 @@
 Last but not least this chapter ships the API reference in detail. Despite explaining every single method its arguments and return value, you will also find some tips on some pages.
 
 ## Top-Level API
-* [`createRenderer([config, initialState])`](api/createRenderer.md)
+* [`createRenderer([config])`](api/createRenderer.md)
 * [`render(renderer, mountNode)`](api/render.md)
 * [`combineRules(...rules)`](api/combineRules.md)
 * [`enhance(...enhancers)`](api/enhance.md)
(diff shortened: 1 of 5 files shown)


johanneslumpe:
@dustin-H We discussed going with #2, but in a less hacky way, by providing a proper way of hydrating the renderer. I'm sure there is room for optimizing what gets sent down the wire, and we might not need to send everything if we can derive the complete state from several sub-states.

That said: gzip will most likely compress things a lot so I wouldn't worry about it too much for now.

I will leave the merge-decision for this one to @rofrischmann.

Issue #70, opened 2016-08-02
Theming


rofrischmann:
There should be an easy way to do theming.

We could use an enhancer to add a theme as second parameter of each rendering method.
Yet one could also theme using the props or a simple Theme singleton as well.

### Theming Enhancer
```javascript
const theme = { 
  PRIMARY: 'red', 
  SECONDARY: 'blue' 
}

const renderer = createRenderer({
  enhancers: [ theming(theme) ]
})

const rule = (props, theme) => ({
  fontSize: props.fontSize,
  color: props.color || theme.PRIMARY,
  backgroundColor: theme.SECONDARY
})

renderer.renderRule(rule, { fontSize: '15px' }) // => 'c0 c0-foo'
renderer.renderRule(rule, { fontSize: '20px', color: 'green' }) // => 'c0 c0-bar'
```
```CSS
.c0 {
  color: red;
  background-color: blue
}

.c0-foo {
  font-size: 15px
}

.c0-bar {
  font-size: 20px;
  color: green;
}
```

#### Base Styles
```javascript
const theme = { 
  PRIMARY: 'red', 
  SECONDARY: 'blue' ,
  styles: {
    button: { 
      color: 'blue', 
      lineHeight: '20px', 
      padding: '10px' 
    }
  }
}

const rule = (props, theme) => ({
  ...theme.styles.button,
  fontSize: props.fontSize,
  backgroundColor: theme.SECONDARY
})
```

MicheleBertoli:
Well done, I love the solution.

Did you already think about a function to change the theme?

johanneslumpe:
@rofrischmann @MicheleBertoli 
What do you think about this:

Instead of returning an enhancer directly, `theming(theme)` could return an object with a shape like `{ enhancer, changeTheme }` . `changeTheme` would be a function which takes a new theme as a parameter to allow on the fly theme changes. That function could then easily be passed down via context in react for example (or via any other way in other frameworks).

The only downside I see here (in a react context) is that this change wouldn't cause a re-render because it's outside of react. This could partially be mitigated by keeping the theme object in a redux store for example, but it still might not force a complete re-render.

rofrischmann:
I thought about adding `renderer.updateTheme(theme)` to the enhancer to trigger theme updates, but I am yet not sure how to rehydrate all styles. One idea would be to just clear all styles and rerender everything (e.g. using `forceUpdate` on the `<Provider />`.
[...]

PR #74, opened 2016-08-05
#70 theming enhancer

rofrischmann:


rofrischmann:
There are acutally better ways to do theming more straightforward. More information coming soon.

PR #75, opened 2016-08-11
Add nested attribute selector support

einarlove:
Support attribute selectors like `[disabled]`alongside the pseudo classes and media queries.

Example
```js
const styles = {
  button: {
     color: 'red',
     ':hover': {
       cursor: 'pointer',
     },
     '[disabled]': {
       cursor: 'default',
     },
  }
}

### DIFF
--- a/modules/plugins/validator.js
+++ b/modules/plugins/validator.js
@@ -2,7 +2,7 @@ function validateStyleObject(style, logInvalid, deleteInvalid) {
   Object.keys(style).forEach(property => {
     const value = style[property]
     if (value instanceof Object && !Array.isArray(value)) {
-      if (property.indexOf(':') === 0 || property.substr(0, 6) === '@media') {
+      if (/^(@media|:|\[)/.test(property)) {
         validateStyleObject(value, logInvalid, deleteInvalid)
       } else {
         if (deleteInvalid) {

### CODE REVIEW

**Comment by rofrischmann** [comment]
Hey @einarlove Thanks for all your help! I really like what you did there. Currently I am unable to push new updates and review code as my Macbook is still recovering from the last beer attack 😅 
Yet I was think about adding sth. like `&` to be able to recreate any CSS selector. e.g.

```javascript
const styles = {
  button: {
     color: 'red',
     ':hover': {
       cursor: 'pointer',
     },
     '&[disabled]': {
       cursor: 'default',
     },
     '& > .child': { 
      // ...
     }
  }
}
```
I would really love if you could just update your PR to implement this behaviour. Will merge and release as soon as possible!

einarlove:
Get yourself a [non spillable cup](https://www.youtube.com/watch?v=8CTxbQ6cdUw) next time 😛.

I'm having trouble understanding why you'd want this syntax. I think it should either go all in on `&:` selector, or not at all. I'll explain.

In CSS you append  pseudo-classes, pseudo-elements, and attributes on the selector
```css
.class[attr="value"]:hover::after { … }
```
You suggest having different ways to declare them in your API.
```js
const styles = {
  button: {
    ':hover': { color: 'blue' },
    '&[disabled]': { color: 'red' },
    '::after': { content: '$' }
  }
}
```
Instead of treating them the same like in [JSS-Nested](https://github.com/cssinjs/jss-nested) which has a consistent approach.
```js
const styles = {
  button: {
    ':hover': { color: 'blue' },
    '[disabled]': { color: 'red' },
    '::after': { content: '$' },
  }
  // or
  button: {
    '&:hover': { color: 'blue' },
    '&[disabled]': { color: 'red' },
    '&::after': { content: '$' },
  }
}
```

You also mentioned that you want support for nested selectors (with global classnames). I'm not sure how it should work in nesting, globals and etc.

My suggestions is to stick with one approach of appending selectors, and do nesting support as a separate issue/pull-request since it requires more work and frankly is a separate feature.

rofrischmann:
I get what you're telling there. Though I am not sure how to do it. Pseudos have and will always be a core feature and should therefore be written with little overhead (without `&` which I always miss on the keyboard anyways).

I consider nested / other selectors as some kind of unsexy hack. But maybe we can just go with the attribute selector you proposed and actually ship another syntax for other selectors at all. 

I would merge tomorrow after doing some brainstorming again :P Didnt check the project for some days actually

PR #89, opened 2016-10-14
Classes are named after their rule when possible

Brian-Gaffney:
Instead of prefixing each rule with 'c' the function name of the rule itself is used when possible.
Uses `function.name` to determine the rule name.
Example `.menuBar_123`.

Makes it easier to figure out which DOM elements and styles belong to which component and which rule.
I've tested this in a React project for a few days and it made development much easier.
It does require rules to be descriptively named for maximum benefit. 

Downside is the CSS and classnames are potentially much longer.
The feature could be hidden behind a config variable passed to `createRenderer` if neccessary.

Browser support for `function.name` - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name#Browser_compatibility


rofrischmann:
Hey Brian! this is actually a great idea. How about only adding this in development and using `c` + ID in production?

Brian-Gaffney:
Hey Robin.
Happy to add that.

How about a config passed to `createRenderer`. Then users can turn it on or off as need in server or browser environments.

`createRenderer({ prettyClassnames: true })`?

rofrischmann:
Let me think about that :P Not sure if I want to introduce a new config key, but might be worth it. But we would call it `prettySelectors` I guess so we can also do it with keyframes.

Also, we should still enforce the `c`+ID style in production or ppl might accidently ship with pretty selectors in production as they forget to remove the configuration.
Could also call it `debugMode` which adds flexibility to add more features of this kind later on.

Brian-Gaffney:
I've updated the PR to use a `prettySelectors` config and an env check to determine whether or not to emit pretty selectors.

Let me know what you think.
If the config is fine by you I'll update the docs as well.
If not I'll just use the env check and remove the config.

modules/createRenderer.js
```diff
@@ -15,12 +15,18 @@ import cssifyObject from './utils/cssifyObject'
  * @return {Object} new renderer instance
  */
 export default function createRenderer(config = { }) {
+  const prettySelectors = config.prettySelectors || false
```
rofrischmann: You can remove this and use `config.prettySelectors` directly as if undefined it is falsy by default

modules/createRenderer.js
```diff
@@ -61,7 +67,10 @@ export default function createRenderer(config = { }) {
 
       // uses the reference ID and the props to generate an unique className
       const ruleId = renderer.ids.indexOf(rule)
-      const className = 'c' + ruleId + generatePropsReference(props)
+
+      const classNamePrefix = this && this.prettySelectors && rule.name ? rule.name + '_' : 'c'
```
rofrischmann: Try not to use `this` explicitly, but rather use `renderer.prettySelectors && rule.name` :P will be fine

modules/createRenderer.js
rofrischmann: You can remove this and use config.prettySelectors directly as if undefined it is falsy by default.

Brian-Gaffney:
I've made those changes you suggested and I've updated the docs.

I also committed changes to `packages/fela/dist/fela.js` and `packages/fela/dist/fela.min.js`.
Not sure if that is the right thing to do.

rofrischmann:
Amazing. Thank you! You would rather not push the dist files, but its not a problem either :P

PR #103, opened 2016-11-15
Component Theming

rofrischmann:


steida:
Wow, that's exactly I need. How can I help you?

rofrischmann:
@steida I guess right now its only about testing the first implementation if its worth using. Then we'd need to write proper documentation. You might try out the ThemeProvider implemented here. (Be sure to also use the `createComponent` and `connect` from this branch as well, otherwise it wont work)

PR #134, opened 2016-11-27
Close #129

### LINKED ISSUE #129: flowtype
Let's use flowtype. It's really easy to adopt it gradually. Just add flow-bin into package.json, and enable it in https://nuclide.io/.

Nothing will be broken.

Then, add `/* @flow weak */` in every visited file once edited. It will not sucks, just help. Later we can improve typing by removing `weak` etc.

It's really that good.

steida:
Even flow weak helps us a lot. Autocomplete, type checking, missing
modules detection, etc.

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,0 +1,9 @@
+[ignore]
+
+# Disable module with broken flow types.
+.*/node_modules/babylon/*.*
+.*/node_modules/esformatter/*.*
+.*/node_modules/config-chain/*.*
+
+[version]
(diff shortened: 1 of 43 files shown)


rofrischmann:
Am going to check and merge tomorrow.
Going to bed now, see you :P

steida:
Sure, take your time. Thank you for your work :-)

steida:
Remember to install nuclide and enable flow-bin to check that beauty ;)

PR #150, opened 2016-11-30
Fix incorrect font format

johanneslumpe:
It's `.eot`, not `.eof`.

ping @rofrischmann  :D


rofrischmann:
Ghahaa :D Seems like this type has never been used before

PR #168, opened 2016-12-26
Fix variable name conflict in React usage docs

rileyjshaw:
In the example, we want to pass the outer `title` rule into `renderer.renderRule`. Before this commit, that name was overwritten by one of the destructured props.


rofrischmann:
Thank you!

PR #193, opened 2017-01-31
feat(react-fela): Use `Themed<displayName>` for connected components

Craga89:
Currently, any `HoC`'d component that is wrapped using `connect` gains the same `displayName` (or similar) as it's own. This can be a little confusing when viewing the React DOM using the Developer Tools:

React DevTools

This feature simply preprends the name with ` Themed` prefix for easier visual distinctions.

### DIFF
--- a/modules/bindings/react/connect.js
+++ b/modules/bindings/react/connect.js
@@ -3,13 +3,13 @@
 import React, { Component, PropTypes } from 'react'
 
 const generateDisplayName = (Comp) => {
-  const displayName = Comp.displayName || Comp.name;
+  const displayName = Comp.displayName || Comp.name
   if (displayName) {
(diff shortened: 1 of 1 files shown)


rofrischmann:
Hey @Craga89,
Thanks for your effort! But why would you want to use the "base" component and the styled "base" component separately? Still I think it's a nice idea to have such a prefixing, but maybe we could use "Fela" as a prefix and not just "Themed"? Seems more clear to me.

Craga89:
IT's not so much that you'd use the "unstyled" component separately, and more that it it's a little unclear immediately when looking over the DevTools React Tree. I think a `Fela` prefix would be better now that you mention it; I'll change it over.

rofrischmann:
Thanks a lot :)

Issue #195, opened 2017-02-08
Monolithic mode?


tiagojsalmeida:
**Version**: 3.0.x
**Environment**: Both
**Type**: Feature

-------------------
#### Description
Atm you guys only support atomic classnames.
Would it be possible to implement an monolithic mode?

#### Proposal

**Code Example**:
```javascript
createRenderer({
   mode: 'monolithic'
})
```

rofrischmann:
Not sure about that. I don't see any advantages actually. But it might be possible to just have an alternative renderer. The React Native renderer e.g. is also a totally different renderer.

tiagojsalmeida:
@rofrischmann In our case it would be ideal to use the monotithic version on development as it's just easy to change styles.
On our implementation we made our own render method as we don't want to use `sheet.insertRule` for multiple reasons (specially because of third-party applications).

I would totally help. Could you point me on the right direction?

rofrischmann:
You could basically take the existing renderer and simply replace renderRule. I may give you a list of "things" which you need to do in order to achieve correct classnames etc.

rofrischmann:
Added with #200

PR #200, opened 2017-02-13
Feature/monolithic

tiagojsalmeida:
## Issue
https://github.com/rofrischmann/fela/issues/195

## Description
Added `monolithic` enhancer

## Usage
```
import monolithic from 'fela-monolithic'

createRenderer({
   enhancers: [ monolithic() ]
})
```


rofrischmann:
Wow. That was fast! Nice one. I will review it as soon as possible :) Could you perhaps update the README and package.json etc. to match the name (instead of `fela-beautifier`)

tiagojsalmeida:
My mistake @rofrischmann .
The caching system is a bit different that yours since it's not atomic style caching

rofrischmann:
Yeah I know that. I've written monolithic renderers myself in the past :) Btw. We should have a notice in the README teaching the differences in usage (e.g. shorthand-longhands - you can use them in monolithic mode without any problem, but it might break in atomic mode)

tiagojsalmeida:
@rofrischmann I added a few entries on the docs. Where would you add more information?

docs/advanced/Enhancers.md
```diff
@@ -42,7 +42,21 @@ function beautifier(renderer) {
...
 
 const config = {
-  plugins: [ beautifier ]
+  enhancers: [ beautifier ]
+}
+
+const renderer = createRenderer(config)
+```
+
+### Monolithic class names
```
rofrischmann: Guess we should remove this one here and rather rename "Hot Reloading" to "Developer Experience" and have both "Hot Reloading" and "Monolithic CSS classes" as headers.

docs/introduction/Ecosystem.md
```diff
@@ -43,6 +43,7 @@ Many plugins and enhancers are already included in the [main repository](https:/
 * [fela-layout-debugger*](https://github.com/rofrischmann/fela/tree/master/packages/fela-layout-debugger) - Adds colored outlines or backgroundColors to debug layouts
 * [fela-logger*](https://github.com/rofrischmann/fela/tree/master/packages/fela-logger) - Logs every rendered change output
 * [fela-perf*](https://github.com/rofrischmann/fela/tree/master/packages/fela-perf) - Logs performance information (time elapsed while rendering)
+* [fela-monolithic*](https://github.com/rofrischmann/fela/tree/master/packages/fela-monolithic) - Renders monolithic classnames
```
rofrischmann: I am not sure if we explicitly need to add the development-only flag? Ppl might also want to use it in production (although I am not sure why^^)

tiagojsalmeida: It's def production ready :)
Someone might use it just to make sure they don't get surprises when using atomic mode in prod and monolithic in dev.

docs/introduction/Ecosystem.md
```diff
@@ -43,13 +43,14 @@ Many plugins and enhancers are already included in the [main repository](https:/
...
 * [fela-perf*](https://github.com/rofrischmann/fela/tree/master/packages/fela-perf) - Logs performance information (time elapsed while rendering)
+* [fela-monolithic](https://github.com/rofrischmann/fela/tree/master/packages/fela-monolithic) - Renders monolithic class names
 
 ### Components
 * [kilvin](https://github.com/rofrischmann/kilvin): Primitive React Layout Components
 
 ### Tools
 * [fela-tools](https://github.com/https://github.com/rofrischmann/fela/tree/master/packages/fela-tools): Useful tools for working with Fela
-* [fela-styles-connector](https://github.com/dustin-H/fela-styles-connector): Simplified react-fela `connect` with autobound styles
+* [fela-styles-connector](https://github.com/dustin-H/fela-styles-connector): Simplified react-fela `connecgt` with autobound styles
```
rofrischmann: woops :P

tiagojsalmeida:
@rofrischmann at the moment a new class name will be generated every time the props change (if props are being used).
I was thinking of actually adding a unique classname for the props.

Per example:
```
const rule = props => ({
      color: 'red',
      background: props.color,
    })
``` 
This would generate 2 classes. One for `.fela-fn31m{color: red}` and one `.fela-45jkdm{background:'blue'}` per example. So these ruleset would have `fela-fn31m fela-45jkdm` class names and only the last would change when props change.

Over-engineering?

rofrischmann:
This is how Fela used to "work" in the past. But the problem is the diffing: it's pretty slow and doesn't work fine with nested props. (or how would you achieve that?)

tiagojsalmeida:
Well, at the moment i'm `json.stringifly` the whole styles to get a proper unique ID.
I would have to get all the properties with prop values so that could get expensive. But it should be faster when the prop changes and generates only one small class.

rofrischmann:
How would you get all properties with prop values? That's not possible :P

tiagojsalmeida:
all the `undefined` ones hahahah

tiagojsalmeida:
hmm they are computed already, damn..

rofrischmann:
Yeah exactly. I used to do `rule({})` to get all the `undefined`. But this doesn't work with nested props e.g. `color: props.theme.color` as it throws `color` is not a property of `undefined` :P

And the whole diffing stuff is just extremely slow (compared to the atomic mode)

tiagojsalmeida:
I think there's a way @rofrischmann 
```
const props = new Proxy({},{get() {
  return props;
}})
```
We could play around with the Proxy object. So all the props will have a proxy object basically and we could just filter them.

rofrischmann:
Just checked the code. Really looks fine to me! Ill merge it and we can then further improve it with your Proxy idea?

tiagojsalmeida:
Awesome @rofrischmann :)
You will need to run `npm publish` though right?
Also, are the bundle version fine?

rofrischmann:
Yeah, I'll publish it as soon as possible. Got a small improvement which I also want to include. But I'll do it this evening. (German time here: 3.48pm right now) The bundles should be fine yeah! They'll be generated automatically on release.

tiagojsalmeida:
Sure, no problem!

Issue #201, opened 2017-02-16
Improve Monolothic enhancer with custom class names


tajo:
**Version**: 4.2.3
**Environment**: Both
**Type**: Feature

-------------------
#### Description
**Fela is awesome** but there are some projects that can't use JavaScript (typically 3rd party things as Zendesk, forum...)  and would like to use the same styles as our Fela based project. The problem is that all generated classes are random and atomic. 

The new monolithic enhancer (https://github.com/rofrischmann/fela/pull/200) partially solves this but the class names are still not human readable and they are based on properties/values, so they change with every rule update.

#### Proposal
We could introduce an extra style declaration called `classname`. Developer could put there an arbitrary string (name). (In our project, we would have a HOC that would assign React displayName into it.)

 Monolithic's [`generateClassName()`](https://github.com/rofrischmann/fela/blob/master/modules/enhancers/monolithic.js#L81) would get styles (before stringify), try to find a property `classname`, use it as a className instead and deleted `classname` declaration (since it's not valid CSS). In case of no `className`, the current logic would be preserved.

The output would be pretty useful for external usage. Any thoughts?
I'm happy to create a PR.

rofrischmann:
You could of course use `renderer.renderStatic` for named classes, but it should not be abused for stuff other than global resets and third-party styles. (if you'd only need the `classname` hack for a minimal set of styles, `renderStatic` should be fine though). 

Also keep in mind, that the monolithic renderer is not at all as fast as the atomic one and especially built for debugging (you can use it without any problems in prod too, of course, but that's not the focus of Fela :P)

It would be good to know, how often you'd want to use this `classname` hack. I'm not sure if it is a good idea to do it at all, if your whole code base in built with it. 

If none of the above makes sense, then you might get a POC ready, so I can test and review the idea in depth. I am willing to accept any sort of PR as long as it is not part of the core but rather can be opt-in through a plugin/enhancer.

If there're still questions, feel free to check the Fela Gitter chat or catch me private on Gitter :P

rofrischmann:
ll release this today :)

PR #204, opened 2017-02-21
Custom classnames for Monolithic

tajo:
Implements https://github.com/rofrischmann/fela/issues/201. 

I might publish the script that will generate the whole BEM like stylesheet using this feature. Not sure if that should be a part of fela since it will be React specific and call the renderer repeatedly. I hope you got the idea. Anyway, that will take longer, since it's not our top priority. We have to transition first. This PR just opens the door. :)

### DIFF
--- a/packages/fela-monolithic/README.md
+++ b/packages/fela-monolithic/README.md
@@ -28,7 +28,32 @@ import monolithic from 'fela-monolithic'
 const renderer = createRenderer({
   enhancers: [ monolithic() ]
 })
+
+const rule = () => ({
+  className: 'custom',
+  color: 'red'
+})
+
+renderer.renderRule(rule)
 ```
+
+outputs
+
+```css
+.fela-custom {
+  color: red
+}
+```
+
+if `className` property is not used, the output will be
+
+```css
+.fela-137u7ef {
+  color: red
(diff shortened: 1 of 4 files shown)


rofrischmann:
Guess we could somehow link that script, as it seems quite useful :P Anyways, I am happy to review your PR asap. Thanks!

packages/fela-monolithic/README.md
```diff
@@ -5,7 +5,7 @@
 
 The monolithic enhancer will use unique class names instead of atomic ones.
 These generated class names are not re-usable like the atomic design but allows you to debug and modify styles with ease.
-Every ruleset will have it's own unique class - this means that a new class will be generated if you are using props and they change.
+Every ruleset will have it's own unique class - this means that a new class will be generated if you are using props and they change. If you want to fix the class name, you can add property `className` into your rule set. This can be useful if you want to generate an external stylesheet that's human readable, re-usable and can be used on non-JS projects.
```
rofrischmann: Would be cool to add an example to the Usage section here? So ppl understand why and what happens :P

tajo: Yea, updated. :)

Issue #209, opened 2017-02-25
More detailed Typescript definition


PavelPZ:
I prepared https://github.com/PavelPZ/fireb/blob/master/typings/fela.d.ts for my fela testing. It contains more detailed version of your index.d.ts. 

It is not tested yet (I need to install fela by JSPM first). Are you interested in such activity? After some testing I can prepare better (commented etc.) version.

rofrischmann:
Sure, I am really happy with any contribution for TypeScript as I am not using it my own!

PavelPZ:
Ok, great, see https://github.com/rofrischmann/fela/pull/210.

rofrischmann:
merged

PavelPZ:
I am sorry, I forgot correct ```string``` return value for following IRenderer methods:
```
    renderRule(rule: TRule, props: TRuleProps): string;
    renderKeyframe(keyFrame: TKeyFrame, props: TRuleProps): string;
```
I modified my https://github.com/PavelPZ/fela fork. Could you please merge once more?

PR #210, opened 2017-02-27
More detailed index.d.ts

PavelPZ:


packages/fela/index.d.ts
```diff
@@ -1,13 +1,44 @@
...
+  type TEnhancer = (renderer: IRenderer) => IRenderer; //http://fela.js.org/docs/advanced/Enhancers.html
+
+  interface IRenderer {
+    renderRule(rule: TRule, props: TRuleProps): void;
+    renderKeyframe(keyFrame: TKeyFrame, props: TRuleProps): void;
+    renderFont(family: string, files: Array<string>, props: TRuleProps): void;
+    renderStatic(style: string, selector?: string): void;
     renderToString(): string;
-    subscribe(callback: Function): any;
+    subscribe(event: (msg: string) => void): { unsubscribe: () => void; }
```
rofrischmann: Looks great!
Just the subscribe seems wrong. It gets called with change objects, not a string. 
e.g. `{type: 1, selector: '.a', declaration: 'color:red' }`

PavelPZ: Sorry, I did not mention example on http://fela.js.org/docs/api/fela/Renderer.html#subscribelistener. I can see there: ```{ type: 'rule', style: 'font-size:12px', selector: 'a', media: '' }```. 

The only question is what to choose: ```declaration: 'color:red'``` or ```style: 'font-size:12px'``` ?

TS definition could be:
```
interface ISubscribeMessage {
    type: string;
    selector: string;
    style: string; //??? OR declaration:string; ???
    media: string;
}
...
subscribe(event: (msg: ISubscribeMessage) => void): { unsubscribe: () => void; }
```

rofrischmann: Gotcha, Seems like the docs are out-dated kinda.

Consider the following cases:

1: Rules & Static style (object)
```javascript
// type === 1
{ selector: string, declaration: string, media: string, static?: boolean, type: number }
```

2. Static Style (string)
```javascript
// type === 4
{ type number, css: string }
```

3. Keyframes



export const RULE_TYPE = 1
export const KEYFRAME_TYPE = 2
export const FONT_TYPE = 3
export const STATIC_TYPE = 4
export const CLEAR_TYPE = 5

rofrischmann:
Great! Thanks a lot, will merge and release soon.

PavelPZ:
There are fix already, see https://github.com/rofrischmann/fela/issues/209#issuecomment-284192384

PR #211, opened 2017-02-28
Demo JS executes before HTML is parsed

electerious:
The demo currently throws an error because the script executes before it can get the `#welcome` element. This error might not appear when using a server with a network delay. Every delay causes the script to be executed later => The error doesn't appear.

```
TypeError: null is not an object (evaluating 'document.getElementById('welcome').className = className')
```

I moved the script to the end of the body. The HTML is now parsed and ready when the script executes.


rofrischmann:
Wooops. Thank you!
I just saw the demo still uses version 1.1.0 😄 8 months old stuff

PR #217, opened 2017-03-09
Throw an error when createComponent() is used with renderer missing

tajo:
We should give users a nice error if they try to use fela based components (components that are using `createComponent` inside) but the renderer is not present in the context.

**Right now it throws something like this:**

```
 TypeError: Cannot read property 'renderRule' of undefined
```

**New message:**

```
createComponent() can't render styles for the component 'Button' without
Fela renderer in the context. Missing react-fela's <Provider /> at the app root?
```

The fact that a component is using fela can be completely hidden, so this can avoid some confusion.


rofrischmann:
Excelent :)

PR #224, opened 2017-03-19
TypeScript fixes

fongandrew:
Fixes the following issues with the TypeScript type def:

* Fixes TypeScript complaining about React not having a default import

* Allows `renderStatic` to be called with a style object instead of a string


rofrischmann:
Thanks a lot for your PRs :)

PR #228, opened 2017-03-22
render children[0] in preact's providers

efflam:
render children[0] in preact's providers because children is always an array in preact 
[https://preactjs.com/guide/differences-to-reac](https://preactjs.com/guide/differences-to-react)


rofrischmann:
Is there actually a problem with the children being an array btw?

PR #229, opened 2017-03-23
Add TypeScript typings to package.json

fongandrew:
Typings were not included in last publish because I forgot to add them to the package.json file =P


rofrischmann:
Oh my bad, forgot to check that either.

PR #231, opened 2017-03-24
Fixed media queries inside validator

tiagojsalmeida:
Validate media queries as nested property.


rofrischmann:
Thanks a lot :)

tiagojsalmeida:
Thank you for always improving fela @rofrischmann !

tiagojsalmeida:
@rofrischmann let me know when it's published.

tiagojsalmeida:
@rofrischmann would it be possible to make a release?

PR #235, opened 2017-03-29
Docs update for react native

JRobber:
Simple update to the React-Native guide to make differences between it and React a little more clear.


rofrischmann:
Thanks!

Issue #240, opened 2017-04-10
fela-monolithic can reuse the same className for different styles


AugustinLF:
**Version**: 4.3.2 (tested on master)
**Environment**: Both
**Type**: Bug

-------------------
#### Description
When using [fela-monolithic](https://github.com/rofrischmann/fela/tree/master/packages/fela-monolithic), if you use the same className for two different sets of rules (that can easily happen if you set the className at the root of the style function when using [fela-plugin-extend](https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-extend)), fela will render the same className for each case, consequently ignoring the styles beyond the first call to the renderer.

While this might to some extent logical, the user should set different names for different styles, it's not mentioned in the doc, and I did spend quite some time to figure the cause of the issue :sweat_smile:

#### Expected Behavior
```javascript
    const rule1 = () => ({
      className: 'custom',
      color: 'red'
    })
    const rule2 = () => ({
      className: 'custom',
      color: 'green'
    })
    const renderer = createRenderer(options)
    const className1 = renderer.renderRule(rule1)
    const className2 = renderer.renderRule(rule2)
    // className1 !== className1
```

#### Actual Behavior
```javascript
    // className1 === className1
```

While this example is obviously a user's mistake (but still, should it fail silently?), this gets less straightforward when using conditions. For instance, this also happens when using extend:
```javascript
const rule = (prop) => ({
    className: 'custom',
    color: 'green',
    extend: {
        condition: prop.red === true,
        style: {color: 'red'},
     }
})
```
If two components depend on rule, they'll end up with the style of the first one created.

I opened a PR #241 with the failing test-case.

rofrischmann:
I am just working on some stuff including this. Wouldn't it be better to throw a warning and go on? I don't think we should "handle" this, because the user explicitely wants `custom` to be the className right?

AugustinLF:
Hmmm, I don't really know. A warning could indeed be sufficient. The confusing part was the error, it does make sense to prevent such thing.

However, what would then be the solution for people needing monolithic and relying on conditionals in their rules? For me it's ok, I wanted monolithic for a better debugging experience, so I just dropped it^^ Another solution would be to add a suffix to such classes.

rofrischmann:
Well we could have sth. added to fela-tools for that. e.g. where the second parameter adds "variations", "options" however you want to call them^^

```javascript 
const rule = (props) => ({
    className: generateUniqueClassName('custom', props),
[...]

PR #241, opened 2017-04-10
Add test for wrong reuse of classnames with monolithic

AugustinLF:
This failing test illustrates the issue raised in #240. I'll gladly solve this issue, and submit the corresponding patch if you're interested. If so, do you have any advices? I identified the source of the problem [line 121 of monolithic](https://github.com/rofrischmann/fela/blob/master/modules/enhancers/monolithic.js#L121), where we just check if the className is already used, without checking if the styles correspond.

Naively, I would add a check for the content of the style, and add a hash after the generated name, but I'm not sure it would be the most efficient (especially performance wise). What do you think of that?

### DIFF
--- a/modules/enhancers/__tests__/monolithic-test.js
+++ b/modules/enhancers/__tests__/monolithic-test.js
@@ -152,4 +152,20 @@ describe('Monolithic enhancer', () => {
 
     expect(renderer.rules).toEqual('.fela-custom{color:red}')
   })
+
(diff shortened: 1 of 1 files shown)


tajo:
Seems definitely like a bug.

rofrischmann:
@tajo Can you check that? I'm not really into the monolithic renderer at all :< I could bring up my old monolithic renderer to compare the mechanisms if that helps.

PR #245, opened 2017-04-19
React 15.5 update: Use prop-types instead React.PropTypes

tajo:
This PR will make react-fela stop throwing warnings with React 15.5 because React.PropTypes is deprecated. `prop-types` should be a dependency of react-fela now.

Non-related question: Do you know https://lernajs.io/? Is there some reason why you don't use it for fela?


rofrischmann:
As for the lerna question. I've seen that and actually love it, but had no time (and no benefit as everything works fine now) to switch. Although, I would love to use Lerna :P

PR #238, opened 2017-05-02
Embed Fonts/Keyframes in Rules

rofrischmann:
**Version**: 4.x.x
**Environment**: Both
**Type**: Feature

-------------------
#### Description

It's a little bit inconvenient that there is a different workflow to handle fonts. Why not to support them everywhere and automatically render them into separate style node on the background?

#### Proposal

Rules could newly contain `fontFace` property.

**Code Example**:
```javascript
const rules = () => ({
  color: red,
  fontFace: {
    fontFamily: 'Lato',
    src: [ 
      'url('./fonts/Lato.ttf') format(truetype)',
      'url('./fonts/Lato.woff') format(woff)'
    ]
  }
});
```

I am not keen on the specific format but you got the idea. 

**Why this?**

- It would make possible to introduce new fonts when component is used. Now you have to do it at the root of your app globally.
- It's easier to set it up. Well, there is no extra setup.
- You don't need an instace of Fela renderer to render fonts, so you can just nicely abstract the whole Fela setup like [here](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-provider/src/index.js#L40).

### DIFF
--- a/modules/createRenderer.js
+++ b/modules/createRenderer.js
@@ -61,7 +61,7 @@ export default function createRenderer(
 
     renderRule(rule: Function, props: Object = {}): string {
       const processedStyle = processStyleWithPlugins(
-        renderer.plugins,
+        renderer,
         rule(props),
         RULE_TYPE
       )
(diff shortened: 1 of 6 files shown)


tiagojsalmeida:
+1 Awesome!

PR #257, opened 2017-05-10
Validator plugin: fix keyframe object validation

isymchych:
Right now validator plugin prints warning if we use `to` or `from` instead of percentage:

```
validator.js:75  Invalid keyframe property.
              Expected either `to`, `from` or a percentage value between 0 and 100. 
Object {percentage: "from", style: Object}
```


rofrischmann:
great thanks :)

PR #262, opened 2017-05-17
Lerna - initial switch

tajo:
It's a bit awkward and scary to reorganize a project that's not mine but #yolo. :)

## Why lerna?

- it was created exactly for projects as fela (monorepo)
- we can start version packages independently, that should promote more releases
- it handles all logic around symlinking, version bumping, tagging and publishing

The current project structure can be a bit confusing since `/packages` is just an empty shell pointing to `/modules`.  Also, there are bundles and compiled files committed. 

This PR puts everything into `/packages`. I was trying to avoid major changes so I preserved the architecture where everything lives  in the `fela` package. So, we have independent versioning now but since all source code is in the `fela` package and all other packages have `fela` as a dependency, it's still 100% dependent.

The next phase (and PRs) si to start breaking down `fela` package and move the code where it belongs. `fela` should contain only code that's shared. **We need to avoid imports as `fela/utils/foo` and import only what's exported from `packages/*/src/index.js`**. This gives us clean APIs between packages and we can fully rely on semver.

@rofrischmann What should we do with `fela/utils`? It should be reexported by `packages/fela/src/index.js` or maybe moved into [css-in-js-utils](https://github.com/rofrischmann/css-in-js-utils)?

## Two things disappeared

- UMD builds (dist folders). I think they are obsolete now thanks to ES6 modules. But we can add them back if you insist.
- README generation - This should go back. I haven't looked into it yet.

## New scripts

- yarn run bootstrap
- yarn run clean
- yarn run publish

^^ Aliases for lerna commands.

TLDR; **In theory, if you merge this everything should be fine and working.** :innocent:


rofrischmann:
😱😍 cc @thejameskyle

tajo:
Also, `packages/*/src/index.js` are still using commonjs so it's not taking advantage of tree shaking. I didn't want to touch it with this already bloated PR. 

**Ad utils:** So we could inline some of them and reexport the rest? I suspect that most of packages could be completely independent on fela? (It's always nice keep the number of cross dependencies low).

tajo:
Maybe even creating `fela-utils` ?

rofrischmann:
If we do not create any pain with fela-utils, I am fine with that. But I used to go a similar path with react-look 2 years ago and I hated it (although I did not use lerna to handle all the stuff). As long as it doesn't get harder to maintain or breaks anything - do whatever you think will fit the best =)

rofrischmann:
Could you update from master once more? Just added all changes for a clean release :/ Not many files touched though, hope that won't be too much work again.

tajo:
@rofrischmann Done.

PR #250, opened 2017-04-27
Docs: Testing Components

derek-duncan:
Inspired by #247. WIP. Feel free to review/contribute.

PR #264, opened 2017-05-18
Linking to new testing doc

derek-duncan:
Forgot to add links to the testing doc in #250 🙃 
Anywhere else it should be linked at?


rofrischmann:
I guess thats it :) but should thought itd maybe better fit into recipes?

derek-duncan:
That works. I'll the docs over 👍

PR #268, opened 2017-05-23
Docs: Correct package import

brentertz:
This is a simple documentation update to correct the import for `createRenderer` to come from the `fela` package rather than the `react-fela` package.


rofrischmann:
thanks :)

PR #273, opened 2017-05-30
Break fela into packages and fix exports

tajo:
First iteration. Stuff is broken still but the new structure / split is pretty much done.

**Todo:** 
- [x] fix test, lint, flow
- [x] test with `packages/example-react`
- [x] move all examples into `packages/`

**Major release will be needed because:**
- missing dist (umd) builds
- fela-dom now exports { render, renderToString, renderToMarkup }, no `fela-dom/server`
- `add-default-export` is gone, you always have to use named exports
- people could rely on some of these internal `fela/lib/` imports
- better safe than sorry


rofrischmann:
Nice PR! Looking forward to release this :) Is there are reason we can't have fela-dom/server now? (btw. its not even released yet, so it won't break anything :P) fela-dom/server is one of the new features I was working on the last weeks.

tajo:
@rofrischmann We can but the problem is that import like that must be pointed to commonjs build `/lib`. If you keep all exports in one file (like `index.js`), you can use `module` and `main` fields in package.json to let other tools decide what build to import. Also, it creates better contract. It's a bad practice to import internals of the package since library authors don't semver that (I learned this lesson the hard way with React).

Is there some reason why you prefer `fela-dom/server`? I know `react-dom` follows the same logic but they don't provide es module builds. I think it's mostly a band-aid for non-working tree-shaking  in  their case. This way you don't have to add the whole react-dom/server into browser bundle. Anyway, this will be even more relevant once node starts supporting modules which should be soon.

rofrischmann:
Yeah I kinda liked that we have it "separate" with the /server. It's ok for this use case as the /server stuff should never be packaged for the browser anyway - it's pure server-side stuff.

But we can ofc just have a fela-dom-server package or ship it with fela-dom as you proposed to do.

tajo:
@rofrischmann 👍 Should be ready for merge and **major release**. 

I tried to run all examples (after I moved them into packages):

- angular2 - works
- angular2-typescript - some problem with loading `index.d.ts`
- inferno - for some reason it doesn't correctly pass `renderer` to `Provider`
- preact - can't find `h()`
- react - works (updated to webpack2)
- react-native - not sure, never used native before... it says something about packager
- vanilla - works but it's using old UMD build

I tried them also with the previous version (not using the latest deps from this PR) and had same problems with angular2-typescript and preact. Inferno surprisingly works (no errors), however it doesn't render its proper content. React-native works, so it breaks with the latest master and the move.

The good thing about putting these examples into `/packages` is that they will never go out of date with fela so we can keep eyes on them. I think we should fix them in different PRs since this one is already enormous and it's not entirely related. **The module and commonjs exports seem to be fine, that was the main thing I was looking for**.

tajo:
^^ **Not ready for release.** I was doing more testing with cf-ui and cloudflare apps ... it generates classnames correctly but it doesn't put them into `<style></style>` and no errors. Strange. I need more time to look into it.

PR #281, opened 2017-06-02
Style nodes initialization and duplication fix

tajo:
- `fela-dom/render` should mount all `<style>` nodes and rules if necessary even before the first change event is triggerd.  These cases were not covered:
  - style nodes were not server-side rendered
  - style nodes were server-side rendered but the rules don't match with browser version
- there was a weird bug when `fela-dom/render` was called twice right away (aka `react-fela`'s `<Provider>` renderered from two points of app) - it created all style nodes twice. I believe it's because the list of mounted nodes were kept in `fela-dom/render()` closure and there was some delay between adding nodes into DOM and reading them from DOM. Now, this list of mounted nodes is kept in `renderer` which should be always singleton.

~**Tests**: I'm trying to put together some tests for the style nodes initialization but so I haven't been lucky with JSDom yet. I'll try later again.~ We **still** need better coverage for `fela-dom`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "babel-preset-es2015-rollup": "^1.1.1",
     "babel-preset-react": "^6.5.0",
     "babel-preset-stage-0": "^6.5.0",
+    "beautify": "^0.0.8",
     "codeclimate-test-reporter": "^0.3.1",
     "cross-env": "^1.0.8",
     "css-in-js-utils": "^1.0.3",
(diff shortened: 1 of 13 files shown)


tajo:
@rofrischmann I extracted `getStyleNode` into `fela-tools`.

tajo:
@rofrischmann I've added some snapshot tests.

tajo:
I'm going to add more tests but in different PRs.

rofrischmann:
Awesome! Can't wait to finally release that beast 5.0

PR #282, opened 2017-06-04
Fix format issue in renderFont

as-com:
When I rendered a Base64-encoded TrueType font, fela output something like this:
```css
@font-face{font-family: Blah; src: url(data...) format('ttf')}
```

The font was not working because of the `format('ttf')` part of the output. I have traced it to the checkFontFormat.js file, which did not output [the correct format statement](https://drafts.csswg.org/css-fonts-3/#src-desc) for TTF and EOT fonts. I also used this as an opportunity to make some optimizations to the code.

This pull request:

- changes the format output for .eot fonts from "eot" to "embedded-opentype," as per the standard linked above
- changes the format output for base64-encoded TTF fonts from "ttf" to "truetype"
- adds support for .svgz file extensions
- optimizes the isBase64 function to only check the first 5 characters instead of searching the whole string
- properly parses out the MIME type from the data URI, instead of searching the string several times and risking false output when the data portion happens to contain a MIME type
- properly parses the file extension from file URLs instead of searching the string several times and risking false positives
- adds tests for checking base64-encoded fonts

### DIFF
--- a/packages/fela-tools/src/__tests__/checkFontFormat-test.js
+++ b/packages/fela-tools/src/__tests__/checkFontFormat-test.js
@@ -3,7 +3,7 @@ import checkFontFormat from '../checkFontFormat'
 describe('Checking the font format', () => {
   it('should return the correct format', () => {
     expect(checkFontFormat('foo.ttf')).toEqual('truetype')
-    expect(checkFontFormat('foo.eot')).toEqual('eot')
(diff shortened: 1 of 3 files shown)


rofrischmann:
Thanks for all these improvements! Font rendering has never been a big thing for me within Fela. (y)

PR #290, opened 2017-06-09
docs(connect): fix mismatched prop name

levithomason:
It looks like the prop `size` is mismatched with `fontSize`.  I've normalized them here to `size`.


tajo:
Thanks!

PR #291, opened 2017-06-09
docs(ThemeProvider): remove non-existent flat prop

levithomason:
The [`ThemeProvider`](https://github.com/rofrischmann/fela/blob/master/packages/inferno-fela/src/ThemeProvider.js) does not handle a `flat` prop, yet it is shown in the docs.


rofrischmann:
Thanks for removing :) The flat prop was removed long before and we forgot to check that

PR #292, opened 2017-06-09
Simulation plugin

rofrischmann:


tajo:
Tests are failing. 😮

rofrischmann:
Oh yeah, fixed that. Was not even related to the PR :D

packages/fela-plugin-simulate/src/__tests__/simulate-test.js
```diff
@@ -0,0 +1,64 @@
...
+      simulate()(style, undefined, undefined, {
+        simulate: {
+          '@media (min-height: 300px)': true,
+          ':hover': true
+        }
+      })
+    ).toEqual({
+      width: 20,
+      backgroundColor: 'blue',
+      color: 'blue'
```
tajo: Why is `color: blue` winning over `color: red`? Does the order of props in `const style` matter?

tajo: In that case, I would just delete this test, since it's not necessarily deterministic.

rofrischmann: Yeah, right now it simply iterates all object values and resolves them straight in order. Could later improve that to first resolve medias, then pseudos, etc.

PR #293, opened 2017-06-10
Update preact example and preact-fela

tajo:
- update preact example:
  - strange import of React
  - not using 3rd parameter of render() to overwrite server-side rendered DOM
  - remove fontRenderer
  - use fela-plugin-embedded
  - clean up the project - make it identic to react example
  - update to the latest deps
  - fix webpack config to enable tree shaking
  - exclude node_modules from babel-loader
- update preact to v8 in preact-fela


tajo:
^ I will rebase this once https://github.com/rofrischmann/fela/pull/292 is merged, that will fix the tests

PR #296, opened 2017-06-11
docs(ecosystem): remove inline-style-transformer

levithomason:
Removes `inline-style-transformer` from the ecosystem list, it is deprecated.


rofrischmann:
Thanks for all the small and well-checked patches @levithomason. That's something we really appreciate as it's really hard to keep track of an updated documentation every time.

PR #300, opened 2017-06-13
Fix fela-plugin-embedded nested properties

tajo:
The plugin looks for `animationName` and `fontFace` only on the root level of rules.


manatarms:
Thanks for the quick fix @tajo

PR #303, opened 2017-06-15
Document react provider mountNode prop

AugustinLF:
The provider from `react-fela` takes an optional mountNode prop, that is useful if you need to place fela's style declaration and external ones in a specific order.

### DIFF
--- a/packages/react-fela/README.md
+++ b/packages/react-fela/README.md
@@ -16,7 +16,7 @@ You may alternatively use `npm i --save react-fela`.
 
 
 ## API Reference
-* [`<Provider renderer>`](docs/Provider.md)
+* [`<Provider renderer [mountNode]>`](docs/Provider.md)
 * [`connect(mapStylesToProps)`](docs/connect.md)
 * [`createComponent(rule, [type], [passThroughProps])`](docs/createComponent.md)
 * [`<ThemeProvider theme [overwrite]>`](docs/ThemeProvider.md)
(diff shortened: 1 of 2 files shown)


rofrischmann:
The mountNode prop is actually deprecated as the new fela-dom `render` automatically applies style nodes to the head and reflushes style nodes render with fela-dom `renderToMarkup` or `renderToSheetList`.

If you pass a mountNode, fela uses the old render behavior which is not optimized like the new one. :P 

Maybe we can add a side node for that somewhere.

AugustinLF:
Damn! My problem is the one I explained, when I did update to `fela@5.0`, I did skip the `mountNode`, which totally broke my styles, fela taking precedence over external stylesheets, while it used to not have priority.

I can add a comment to specify that this behaviour should be avoided as much as possible, but that doesn't solve my problem on the long term. Perhaps I should open another issue for this discussion?

rofrischmann:
@augustinLF Are you using server-side rendering? because you can then specify the position of the styles in the head which the fela-dom later adapts. Maybe, for pure client-side use we can introduce some kind of "anchor" element, that will be used to insert the styles after/before.

AugustinLF:
Thanks, using server-side `renderToStylesheet` did the work! I guess I should close this PR then?

Regarding your suggestion, should I perhaps create an issue for that? Even if I don't need it anymore?

rofrischmann:
Could be useful for client-side only usage. Feel free to open that issue :) We can close this one here I guess.

PR #304, opened 2017-06-20
Bugfix: Prevent fallback plugin to mess up fontFace src property

tajo:
This combination of plugins

```js
createRenderer({ plugins: [fallbackValue(), embedded()] })
```

is breaking

```js
{
  fontFace: {
    fontFamily: 'font-name',
    src: [
      'data:application/x-font-woff;charset=utf-8;base64,PASTE-BASE64-HERE'
    ]
  }
}
```

since the fallback plugin changes `src` property.

I've also added a test for base64 into embedded plugin and mentioned the usage in its README (I did that when I thought that embedded is broken but the real cause is fallbackValue).

### DIFF
--- a/packages/fela-plugin-embedded/README.md
+++ b/packages/fela-plugin-embedded/README.md
@@ -49,6 +49,19 @@ const renderer = createRenderer({
 }
 ```
 
+It also supports base64 encoded fonts:
+
+```javascript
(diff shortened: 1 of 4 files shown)


koddsson:
Appreciate the quick fix 👏

PR #306, opened 2017-06-22
Proposal: Proxy to prevent wrong props to pass down

wcastand:
I propose to use Proxy to pass all props by default except the one used in the rules.
The passThroughtProps is still available and allow to use props in the rule and still pass it down the element (and allow retro compatibility).

The main idea behind this is that when we pass props to or component we have 2 use cases:
- pass props for the element(onClick, href, ...)
- pass props to dynamic style (flex, glow, vertical, size, ...)

If we use custom Component and not element(div, span, ...), the rule still applied since what i pass as props and do not use in my rule is probably needed in my custom Component.

I updated the example in example/react-fela with the `Header` which doesn't need to specify 'title' in passThroughtProps since it's not used in the rule.

This solution allow the use of custom component (third party lib, HOC, ...) in a much easier way.
It solves a problem with data pass down customElement that does not concern the user become they are internal logic of third party library([my use case](https://gist.github.com/wcastand/6dd9876d9a82dad3545927df53febcd6))

My main concern about proxy is performance, i read everywhere that proxies are slow. So we should probably make some benchmark or at least test on real app.

Let me know what you think about the idea ;)
Making a PR without discussing it must be a bit quick but the subject was interesting personally so i figured i should give it a go.

### DIFF
--- a/packages/example-react/components/Header.js
+++ b/packages/example-react/components/Header.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { createComponent } from 'react-fela'
+import { createComponentWithProxy } from 'react-fela'
 
 const Header = ({ title, className }) => (
   <div className={className}>
@@ -23,4 +23,4 @@ const rule = () => ({
   }
 })
 
-export default createComponent(rule, Header)
+export default createComponentWithProxy(rule, Header)
--- a/packages/example-react/components/Header.js
+++ b/packages/example-react/components/Header.js
@@ -23,4 +23,4 @@ const rule = () => ({
   }
 })
 
-export default createComponent(rule, Header, ['title'])
+export default createComponent(rule, Header)
--- a/packages/fela-utils/src/resolveUsedProps.js
+++ b/packages/fela-utils/src/resolveUsedProps.js
@@ -1,0 +1,8 @@
+export default function resolveUsedProps(props: Array<string>, src: Object) {
+  const output = []
+  for(let key in src) {
+    if(props.indexOf(key) === -1)
+      output.push(key)
+  }
+  return output
+}
--- a/packages/fela-utils/src/resolveUsedProps.js
+++ b/packages/fela-utils/src/resolveUsedProps.js
@@ -1,8 +1,9 @@
+import objectReduce from './objectReduce'
+
 export default function resolveUsedProps(props: Array<string>, src: Object) {
-  const output = []
-  for(let key in src) {
+  return objectReduce(src, (output, value, key) => {
     if(props.indexOf(key) === -1)
       output.push(key)
-  }
-  return output
+    return output
+  }, [])
 }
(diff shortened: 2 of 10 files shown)


tajo:
Interesting! I have a few questions (I never used proxy before so bare with me):

1. What if I need to pass and access some prop in both rule function and my component?
2. What if my theme is deeply nested (e.g. `theme.color.red` is in my rule function)? Isn't that going to fail with your implementation?
3. Is it going to work for

  ```js
  rule = ({ active }) =>  ({
    color: active ? 'red' : 'blue'
  });
  ```

  or only

  ```js
  rule = ({ myColor }) =>  ({
    color: myColor
  });
  ```

[My solution](https://github.com/cloudflare/cf-ui/blob/master/packages/cf-style-container/src/index.js#L12-L19) for React components: I'm extracting propTypes and adding them to `passThroughProps`. Since we always use propTypes, we almost never have to manually specify `passThroughProps`.

wcastand:
Hi, for the first question passThroughtProps still works so you can specify the props you want to pass even if you use it in your rule :)

Second, it works we don't need to verify nested props and in the proxy I still return the nested value to avoid an error

And finally it works with destructuring because in the end you're still accessing the same value so the call for proxy is the same ;)

Even if it's not good sometimes I don't use proptype for small component or for rapid prototyping/test but that's a really interesting idea.  Maybe it's better and more performant I don't know.

Maybe my implementation need to be an options to activate ?

tajo:
> Maybe my implementation need to be an options to activate ?

I would probably publish it as a new separate API, since it seems a bit magical to just introduce it into createComponent that's very transparent now. Maybe we could combine this with my solution to cover even more cases.

wcastand:
I created createComponentWithProxy to let createComponent as it used to be.

I also updated the doc and create test for createComponentWithProxy.
I copy/paste tests from createComponent to be sure it still works the same and adapt tests where they needed to for the new behavior.

I also add one to be sure that props declare in passThroughProps are still passed even if used in rules.

Let me know if i missed anything :)

rofrischmann:
Finally going to merge this today. Had literally no time before :/ 
Thanks for the great PR already!

wcastand:
You said on twitter you were on vacation so i didn't expect this to be merge sooner.
Take all the time you need ;)

packages/example-react/components/Header.js
```diff
@@ -23,4 +23,4 @@ const rule = () => ({
   }
 })
 
-export default createComponent(rule, Header, ['title'])
```
rofrischmann: You will need to readd the title prop or use the new createComponentFactory or else the Header won't show its title :)

packages/fela-utils/src/resolveUsedProps.js
```diff
@@ -0,0 +1,8 @@
+export default function resolveUsedProps(props: Array<string>, src: Object) {
+  const output = []
+  for(let key in src) {
```
rofrischmann: Would be cool if you could use fela-utils/objectReduce directly

```javascript
return objectReduce(src, (output, value, key) => {
  if(props.indexOf(key) === -1)
    output.push(key)
  }
  return output
}, [])
```

packages/example-react/components/Header.js
wcastand: i forgot to update it, it was suppose to work because it was suppose to use my proxy ahah

PR #307, opened 2017-06-24
Bugfix: Ensure that monolithic className never starts with number

tajo:
Apparently you can't have className as `10h3ujg` (starting with number) unless you escape it. So, it's best to avoid it completely. 

**Simple fix:** If there's no prefix, it just adds `f` (as fela) at the beginning of hashed className. This should be a patch release since people should not rely on classNames anyway.

// This was sooo annoying bug to find since I had no clue that numbers at first position are not allowed. :)


rofrischmann:
Too bad I forgot to think about that :/

tajo:
patch release? plz :)

PR #313, opened 2017-06-30
Pass component's static methods through

### LINKED ISSUE #312: Original component's static methods are not available when using connect
**Version**: 5.0.0
**Environment**: DOM
**Type**: Bug

-------------------
#### Description
As a developer I need access to the original component's static methods

#### Steps to reproduce
1. Make a component with `propTypes`
2. Wrap it in `connect`
3. `console.log` your connected component's `propTypes`

#### Expected Behavior
A connected component should have all original static methods assigned

#### Actual Behavior
A connected component does not have all original static methods assigned

## NOTES
I already fixed it, I'll send the PR in a bit for review

Lepozepo:
Fixes #312

Not too sure if the test I wrote will pass or fail though, couldn’t get tests running locally

### DIFF
--- a/packages/fela/src/bindings/connectFactory.js
+++ b/packages/fela/src/bindings/connectFactory.js
@@ -1,6 +1,7 @@
 /* @flow */
 import { objectReduce } from 'fela-utils'
 import generateDisplayName from './generateDisplayName'
+import hoistStatics from 'hoist-non-react-statics';
 
 export default function connectFactory(
   BaseComponent: any,
@@ -59,7 +60,15 @@ export default function connectFactory(
         }
       }
 
-      return EnhancedComponent
+      // Re-attach static methods
+      EnhancedComponent.childContextTypes = component.childContextTypes;
+      EnhancedComponent.defaultProps = component.defaultProps;
+      EnhancedComponent.getDefaultProps = component.getDefaultProps;
+      EnhancedComponent.mixins = component.mixins;
+      EnhancedComponent.propTypes = component.propTypes;
+      EnhancedComponent.type = component.type;
+
+      return hoistStatics(EnhancedComponent, component)
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -68,5 +68,8 @@
     "react-native": "^0.45.0",
     "react-test-renderer": "^15.5.4",
     "rimraf": "^2.6.1"
+  },
+  "dependencies": {
+    "hoist-non-react-statics": "^2.0.0"
   }
 }
--- a/packages/fela/src/bindings/__tests__/connectFactory-test.js
+++ b/packages/fela/src/bindings/__tests__/connectFactory-test.js
@@ -40,4 +40,36 @@ describe('Connect Factory for bindings', () => {
       toJson(wrapper)
     ]).toMatchSnapshot()
   })
+  it('should expose the components original static methods', () => {
+    const rules = {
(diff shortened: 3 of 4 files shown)


tajo:
@Lepozepo What's the problem with running tests locally?

tajo:
I faced the same problem with `createComponent` and `createComponentFactory` not passing propTypes through. I ended up with forking `createComponentFactory` to fix that. However, my original reason for the fork was to prevent `createComponent(createComponent(...))` since I believe it's an anti-pattern (there are much better ways how to combine styles if necessary).

The problem with your implementation is that you modified `connectFactory` with React specific static properties but `connectFactory` is used by other react like libraries bindings. I think those binding should be somehow refactored so more stuff can be done in packages as react-fela, inferno-fela...

Lepozepo:
@tajo I see what you mean, then we should probably just take what's on hoist-non-statics and remove the blacklist stuff they have going on right? Then all non-react packages can also have access to the original components' statics. Not too sure why my test command didn't run locally >_< do I need to do anything special besides run `test`?

tajo:
@Lepozepo 

```
yarn install
yarn run bootstrap
yarn run build
yarn test
```

rofrischmann:
How the current status on this? Do we have a refactored version that works with all bindings?

Lepozepo:
Not yet, I haven't had a chance to revisit, I'll try to push something out today

rofrischmann:
Ping me if you need any help :) @Lepozepo

rofrischmann:
@tajo I just checked the source here. I guess we can actually merge it as the hoist-non-react-statics package does not require any react-specific code and should work with Preact/Inferno as well.

packages/fela/src/bindings/connectFactory.js
```diff
@@ -59,7 +60,15 @@ export default function connectFactory(
         }
       }
 
-      return EnhancedComponent
+      // Re-attach static methods
+      EnhancedComponent.childContextTypes = component.childContextTypes;
```
tajo: @rofrischmann It looks ok. Are we sure it's legit to hoist also all React statics?

Lepozepo:
Hey guys! Sorry I never got a chance to submit a new PR. For my particular use-case what I actually wanted was the wrapped instance's methods, the problem with doing it this way is that there is still no way to access the original instance and while this does expose methods, they aren't actually attached to the original instance, they're just a part of the new one. I ended up using `refs` and exposing the original component in a static instance variable called `instance`. I'm not sure if that is ok or not. Looking over the way redux `connect` is built, they have a `withRef` option that does something similar (while also hoisting statics).

packages/fela/src/bindings/connectFactory.js
rofrischmann: There're only a few React-specific ones right? Doesn't seem like to much overhead for the other projects right now. We could refactor that later.

rofrischmann:
@Lepozepo That's actually how I'd do it too. Although I've never written a Fela Component that used static methods, because I believe that's probably an anti-pattern. Yet, what would you suggest here? Just add a "recipe" to the docs to help ppl doing it or to actually ship it with the bindings?

Because we could then just close the PR and release the other stuff :P

rofrischmann:
Will close this for now. Don't know if we even need this at all. If so, we should reopen the issue and create an abstract version.

Issue #314, opened 2017-07-03
Elements with class "ad" (and others) are removed by ad blockers


eliseumds:
**Version**: 5.0.2
**Environment**: Both
**Type**: Feature

-------------------
#### Description

As already stated in [this issue from Styletron](https://github.com/rtsao/styletron/issues/139), elements with the class name `ad` get removed by some ad blockers. A few others like `facebook-icon`, `twitter-icon` and `social-icon` are also affected.

#### Steps to reproduce

See this comment: https://github.com/rtsao/styletron/issues/139#issuecomment-305227895

#### Proposal

This could be fixed by a plugin like [fela-monolithic](https://github.com/rofrischmann/fela/tree/master/packages/fela-monolithic) but I suppose a built-in blacklist of classes would be more ideal since pretty much every dev would have to deal with it. To make it more future-proof, user could pass a custom `generateClassName` or a list of class names to avoid.

HugoGiraudel:
As of right now, you can work around this issue by using the `selectorPrefix` option. Use any letter so blacklisted class names from AdBlockers are not being generated.

PR #316, opened 2017-07-07
Fix issue with blacklisted class names from ad blockers

HugoGiraudel:
Hello,

@smartmike and I are facing https://github.com/rofrischmann/fela/issues/314 at N26. 

Not sure if this fix is the way to go as this may cause a conflict with the next generated className? What if this function maintained its internal index with a closure? That would make it possible to just skip to next without risking a conflict.

One other way to make this issue somehow irrelevant (without actually fixing it though…) is to reverse the alphabet. This would enlarge the buffer before facing this problem.

Other possibility that would be much more bulletproof: removing or replacing the `d` from the charset. That’s odd, but that would cover that major edge case.

Other suggestion: adding an extra special character at the end to bypass class checkers.


HugoGiraudel:
Moved to #319.

PR #317, opened 2017-07-07
Fix human readable classname conflict

smartmike:
Related to #316 and #314 – Also not sure if this is the way to go :)


smartmike:
Okay, for sure this isn't working properly, but it's an idea maybe...

rofrischmann:
How exactly does this improve the className readability?

HugoGiraudel:
Moved to #319. Feel free to close, @smartmike.

PR #319, opened 2017-07-12
Added filterClassName renderer option

HugoGiraudel:


packages/fela/src/createRenderer.js
```diff
@@ -238,7 +239,11 @@ export default function createRenderer(
 
             const className =
               renderer.selectorPrefix +
-              generateClassName(++renderer.uniqueRuleIdentifier)
+              generateClassName(
+                ++renderer.uniqueRuleIdentifier,
+                void 0,
```
rofrischmann: I've never seen void 0 ever in action :D

HugoGiraudel: \o/

HugoGiraudel:
Yay!

PR #320, opened 2017-07-15
Name created rule functions in StyleSheet tool

troch:
We use `StyleSheet`  from `fela-tools` and the functions it creates are anonymous. We use the monolithic enhancer for development and therefore some created class names don't contain their rule name. I've added support for a `ruleName` property, as functions can't be named dynamically.

**Before:**
image

**After:**
image


rofrischmann:
Great thanks :)

PR #324, opened 2017-07-21
321 include default props

rofrischmann:
Would want to merge and release this as a new minor as soon as #313 is ready.


tajo:
How about propTypes?

rofrischmann:
It's not about moving the static stuff to the component, but about using the defaultProps for the rule as well. We don't need propTypes here or am I missing something?

PR #326, opened 2017-07-24
Contributing and Code of Conduct

tajo:
Addresses #323 

Also, I thought it would be nice to have a code of conduct.


rofrischmann:
We should also update the Code Formatting part as we are now using Prettier (can be formatted via `yarn format`). So probably providing a link to prettier-atom e.g. would be great. The esformatter stuff can be removed completely. Like that you also added a Code of Conduct :)

tajo:
@rofrischmann done

I've also ran `yarn prettier` which changed some files. Then, I've updated to the latest prettier and re-ran `yarn prettier` which surprisingly didn't change anything.

Issue #331, opened 2017-07-27
innerRef shouldn't be passed down with createComponentProxy


wcastand:
**Version**: 5.x.x
**Environment**: Both
**Type**: Bug

-------------------
#### Description

Right now, `createComponentWithProxy` pass all the props not used in the rules. But the props innerRef shouldn't be passed down to the children because it's used to create a reference in fela.

If your okay with that statement, i can make a PR to filter the innerRef props from the props passed down to the children with `createComponentWithProxy.

We should use that opportunity to be sure they aren't others props that needs to be filtered like innerRef.

#### Steps to reproduce
1. create a Component with createComponentWithProxy
2. pass an innerRef to the component

#### Expected Behavior
```javascript
<MyFelaComponent innerRef="item" title="test" />
// pass down title but not innerRef (because it's used by fela to create ref)

```

#### Actual Behavior
```javascript
<MyFelaComponent innerRef="item" title="test" />
// pass down title and innerRef (which create a warning in React because innerRef is not a valid prop

```

ragurney:
What would be the solution now if I actually want to pass `innerRef` to the child?

PR #332, opened 2017-07-27
add innerRef to be filtered from createComponentWithProxy

wcastand:
add innerRef to be filtered from createComponentWithProxy before passing props down the children.

It might be a good time to add other props that need to be filtered :)


wcastand:
related to #331

rofrischmann:
Would we need to filter the `is` property? It's also createComponent-specific and should not be passed to its children.

rofrischmann:
How comes the test fails? Would like to merge and release :)

Issue #333, opened 2017-08-01
Add support for local system fonts fallback.


aga5tya:
Support local aliases for fonts to fallback. Useful when fonts are already in the available in the system or to avoid cross-origin during development when self hosting fonts on cdn.

PR #335, opened 2017-08-01
feat(renderFont) : Add ability to refer local fonts

aga5tya:
Provides the feature detailed in rofrischmann/fela#333

Adds the localAlias property in options. Accepts a string or array of strings.
> It's actually one reference per font and not per font type. Have extended the options object to include this property.

@rofrischmann Please review.


rofrischmann:
Looks fine!

PR #343, opened 2017-08-07
Add Lusk to list of companies using Fela

arturmuller:
All our apps and (and the website) are now using Fela — time to share it with the world! :)

(BTW, love how easy it is to incrementally adopt it. 💯 )


rofrischmann:
👌 nice to head that :) thank you!

PR #368, opened 2017-09-08
Update Code Sponsor link position

cavneb:
@rofrischmann When the link is found at the bottom of the README, it is more effective in helping funding this project.


rofrischmann:
Thank you. Was not sure were to place it, but: why actually?

Issue #373, opened 2017-09-13
[react-fela] Incomplete createComponent passThroughProps definition.


RafalFilipek:
**Version**: 5.3.0
**Environment**: Both
**Type**: Bug

-------------------
#### Description

Hi!

Right now this code is invalid:

```
const Box = createComponent(style, 'div', props => Object.keys(props));
```

I docs we can read:

> You may also pass a function of props as passThroughProps.

But `PassThroughProps` is defined as `Array<string>`. Also `PassThroughProps` is used only in `createComponent` definitions for HTML tags. Two general definitions for `createComponents` uses `Array<string>` as `passThroughProps` type.

```
export function createComponent<Props, Elem extends HTMLElement>(style: Style<Props>, base: string, passThroughProps?: Array<string>): FelaHtmlComponent<Props, Elem>;
```

So, how can we solve this :) ? 

**First step**: In my understanding, `PassThroughProps` definition should look like:

```diff
- type PassThroughProps = Array<string>;
+ type PassThroughFunction<Props> = (props: Props) => Array<string>
+ type PassThroughProps<Props> = Array<string> | PassThroughFunction<Props>;
```

**Second step**: This change makes `PassThroughProps` a Generic type. While it's used in `FelaInjectedProps` we have to also update this type:

```diff
+ interface FelaInjectedProps<Props> {
- interface FelaInjectedProps {
  // ...
+ passThrough?: PassThroughProps<Props>;
- passThrough?: PassThroughProps;
  // ...
 }
```

**Last step**:  Update all `createComponent` definitions. For example:

```diff
- export function createComponent<Props>(style: Style<Props>, base: "a", passThroughProps?: PassThroughProps): FelaHtmlComponent<Props, HTMLAnchorElement>;
+ export function createComponent<Props>(style: Style<Props>, base: "a", passThroughProps?: PassThroughProps<Props>): FelaHtmlComponent<Props, HTMLAnchorElement>;
```

I will provide PR for that right after we merge #372 :)

Thank you!

PR #374, opened 2017-09-14
Update TypeScript definitions 

RafalFilipek:
Fix for #373


RafalFilipek:
Sorry, I'll have to update this. I forgot to disable prettier and now it looks like whole file has changed.

Issue #379, opened 2017-09-18
Plugin for flow-relative style authoring


TxHawks:
Following the short discussion in [#344 (comment)](https://github.com/rofrischmann/fela/issues/344#issuecomment-3234515140) I published a [library](/txhawks/bidi-css-js) which is based on `rtl-css-js`'s core logic, but enables authoring direction-independent, flow-relative styles. I am planning to publish a fela-plugin wrapping its functionality and was wondering if you'd like it as part of this repository or should I publish it independently.

rofrischmann:
If you want to contribute directly, I'd love to have it added here. You benefit from the build/test/release environment and get auto updates if fela changes :)

Apart from that it sounds awesome what you've done there! Id love to review if required!

TxHawks:
Cool. Heading to bed now, but will make sure to send a pull request tomorrow

PR #381, opened 2017-09-19
bidi plugin

TxHawks:
This PR add a plugin to enable flow-direction independent style authoring, as discussed in #379

### DIFF
--- a/packages/fela-plugin-bidi/src/index.js
+++ b/packages/fela-plugin-bidi/src/index.js
@@ -1,11 +1,6 @@
 /* @flow */
 import transformStyle from 'bidi-css-js'
 
-import type DOMRenderer from '../../../flowtypes/DOMRenderer'
-import type NativeRenderer from '../../../flowtypes/NativeRenderer'
-
-type Type = 1 | 2 | 3 | 4 | 5
-
 export default function bidi(flowDirection: 'ltr' | 'rtl') {
   return (style: Object) => transformStyle(style, flowDirection)
 }
--- a/packages/fela-plugin-bidi/src/index.js
+++ b/packages/fela-plugin-bidi/src/index.js
@@ -1,0 +1,11 @@
+/* @flow */
+import transformStyle from 'bidi-css-js'
+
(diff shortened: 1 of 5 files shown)


packages/fela-plugin-bidi/src/index.js
```diff
@@ -0,0 +1,11 @@
+/* @flow */
+import transformStyle from 'bidi-css-js'
+
+import type DOMRenderer from '../../../flowtypes/DOMRenderer'
+import type NativeRenderer from '../../../flowtypes/NativeRenderer'
+
+type Type = 1 | 2 | 3 | 4 | 5
```
TxHawks: I copied this from `fela-plugin-rtl`, but can't say I understand why the type definitions are there

rofrischmann: Hehe. And I copied that from another plugin that directly uses the renderer. Feels like we can remove those :)

TxHawks:
By the way, there is a bit of a messed up situation here regarding plugin order.
This plugin obviously does not know how to handle arrays of values, such as the ones passed to `fela-plugin-fallback-values`, and therefore has to be passed to the `plugins` array after `fela-plugin-fallback-values`. 
This, on its own, isn't an issue - however, `fela-plugin-prefixer` must be passed to the `plugins` array _before_ `fela-plugin-fallback-values`, meaning rules transformed by `fela-plugin-bidi` will not be prefixed.

I think this probably isn't a problem in the real world, because I don't think any of the values handled by this plugin should ever be prefixed, but there might be something I'm not thinking about.

This is also true for `fela-plugin-rtl`

TxHawks:
@rofrischmann - does this look fine now, it is there anything you'd like me to change?

rofrischmann:
Looks pretty cool, but I am too busy to review that today. Let me check back tomorrow =)

rofrischmann:
Looks great for me! Will merge and release right away :)

TxHawks:
Thanks!

PR #391, opened 2017-09-23
#366 [babel-plugin-fela] Resolve css prop on HTML nodes

### LINKED ISSUE #366: [babel-plugin-fela] Resolve css prop
This feature is basically copied from glamors API and could help to get things done much faster.

rofrischmann:
## ToDo
- [x] Resolve `css` on HTML tags
- [ ] Add `renderer` to contextTypes
- [ ] Tests


TxHawks:
👍  * 100
This is my favorite API in glamorous, emotion, etc.

Will this be subject to the same optimizations that `createComponent` gets from the babel-plugin?

rofrischmann:
The first version won't really benefit from the optimizations, but further versions will do!

TxHawks:
Any estimation on when this is going to land?

rofrischmann:
Pretty sure this will land in November, but not exactly sure how fast I get it done.

rofrischmann:
Damn, this is actually far more complicated than I thought. Looks like it can't be done as simple as glamorous etc. do it, because they all have a global singleton to use, but Fela has the context-sensitive renderer (which actually is a good concept I guess).

PR #397, opened 2017-09-25
fix(connect): Don't pass "theme" prop to wrapped component

### LINKED ISSUE #394: Bug: Theme prop is passed without use of ThemeProvider component
**Version**: 6.0.0
**Environment**:  Both
**Type**: Bug
-------------------
#### Description
Upgraded to fela 6.0.0 on our project (`fela`,`react-fela`,`fela-dom`). After upgrading these packages, a new prop `theme: {}` is present when connecting a component using `connect`. From reading the docs, it is my understanding that this should only occur when a [Theme Provider](https://github.com/rofrischmann/fela/blob/master/packages/react-fela/docs/ThemeProvider.md) is used. This did not occur prior to 6.0.0. If this is an intended change, then this isn't noted in the changelog.

#### Steps to reproduce
1. Have existing code that spreads remaining props over div
2. Upgrade to 6.0.0
3. Observe `Unknown prop  theme on <div> tag. Remove this prop from the element.` error

#### Expected Behavior
To not have the the theme prop present

#### Actual Behavior
The theme prop is present

TxHawks:
closes #394

### DIFF
--- a/packages/fela-bindings/src/__tests__/__snapshots__/connectFactory-test.js.snap
+++ b/packages/fela-bindings/src/__tests__/__snapshots__/connectFactory-test.js.snap
@@ -1,5 +1,48 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`Connect Factory for bindings should not pass through "theme" prop when used without "ThemeProvider" 1`] = `
+Array [
+  "<style>
+    .a {
+        padding: 1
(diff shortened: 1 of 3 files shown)


TxHawks:
@rofrischmann - Is this what you meant?

rofrischmann:
Exactly, that's it :)

PR #403, opened 2017-09-28
[fela-plugin-validator] Adding readme section about plugin ordering

### LINKED ISSUE #401: [fela-plugin-named-media-query] errors
**Version**: 6.0.0 and 5.0.0
**Environment**: Both
**Type**: ~~Bug~~ Docs

-------------------
#### Description
I've been trying to use `fela-plugin-named-media-query` with my app but it appears that I'm always getting a certain error message, even though it does appear to be working behind the scenes:

```
Invalid nested property. Only use nested media queries, pseudo classes, child selectors or &-combinators.
              Maybe you forgot to add a plugin that resolves "bigScreenOnly". {property: "bigScreenOnly", value: {…}}
```

(This error appears multiple times, once for each of the media queries I've create)

I've set it up correctly as far as I can tell:

```
const mediaQueries = {
  bigScreenOnly: `@media (min-width: ${EMediaDimensions.BigScreenStart}px)`,
  tabletOnly: `@media (min-width: ${EMediaDimensions.TabletStart}px) and (max-width: ${EMediaDimensions.BigScreenStart - 1}px)`,
  mobileOnly: `@media (max-width: ${EMediaDimensions.TabletStart - 1}px)`,
  smallMobileOnly: `@media (max-width: ${EMediaDimensions.MobileSmallEnd}px)`,
  tabletUpwards: `@media (min-width: ${EMediaDimensions.TabletStart}px)`,
  tabletDownwards: `@media (max-width: ${EMediaDimensions.BigScreenStart - 1}px)`,
};

const namedMediaQueryPlugin = namedMediaQuery(mediaQueries);

renderer = createRenderer({
      plugins: [
        extend(),
        fallbackValue(),
        unitPlugin,
        validatorPlugin,
        namedMediaQueryPlugin
      ]
    });
```
**Oops... Upon writing this issue I tried to change the order of the plugins and it appears that the error message is gone now. I guess the order is important. Perhaps `fela-plugin-validator` should always come last? What my renderer looks like now:**

```
renderer = createRenderer({
      plugins: [
        namedMediaQueryPlugin,
        extend(),
        fallbackValue(),
        unitPlugin,
        validatorPlugin
      ]
    });
```

From the docs on plugins I see that it says "Order Matters", but there isn't really an indication while using various plugins of what that order should be.

lostpebble:
Updates README.md for validator plugin to inform about ordering.

Fixes #401

### DIFF
--- a/packages/fela-plugin-validator/README.md
+++ b/packages/fela-plugin-validator/README.md
@@ -23,6 +23,11 @@ const renderer = createRenderer({
   plugins: [ validator() ]
 })
 ```
+
+### Plugin ordering
+
+Make sure that you place the validator plugin *at the end* of your plugins array - or else you may get some false error / validation messages.
+
 ### Configuration
 ##### Options
 | Option | Value | Default | Description |


rofrischmann:
Thank you :)

PR #409, opened 2017-09-29
Add yarn.lock to gitignore

vlad-zhukov:
It's outdated and causes problems, `package-lock.json` is already ignored.


alanhogan:
confused — CONTRIBUTING says to run `yarn install` so how is yarn.lock outdated?

rofrischmann:
It is because I am not using yarn atm 😐 might soon switch back if everything works fine again, but for now I guess its fine. Will update the contribution docs anyways. A simple ** run setup should do the magic

vlad-zhukov:
When you decide switching to yarn it makes perfect sense to use workspaces. Have a look at Jest and Babel repos, installation is blazing fast, even tho both are much larger than the fela repo.

rofrischmann:
I've never used that, would you mind setting it up and pointing me to the right directions?

tajo:
I can setup the yarn workspace (it's pretty easy). We switched to it internally and it's great (much faster than lerna).

tajo:
Also, it locks down dependencies of all packages. 👍

vlad-zhukov:
You can also use it along with lerna [just fine](https://github.com/lerna/lerna#--use-workspaces).

Issue #395, opened 2017-09-25
fela-monolithic outputs classes twice


vlad-zhukov:
**Version**: 6.0.x
**Environment**: Both
**Type**: Bug

-------------------
#### Description
Since 6.0 `fela-monolithic` enhancer outputs styles twice when generating with `renderToString` function from `fela-tools` (haven't tried other ways).

Without `fela-monolithic` I get the following:
```css
.a {
  color: red
}
```
But when I add it I get
```css
.rule_Loading__137u7ef {
  color: red
}
.rule_Loading__137u7ef {
  color: red
}
```

If there are more rules it will be like this:
```css
.rule_Loading__137u7ef {
  color: red
}
.rule_Loading__137u7ef:hover {
  color: yellow
}
.rule_Loading__137u7ef {
  color: red
}
.rule_Loading__137u7ef:hover {
  color: yellow
}
```

#### Steps to reproduce
Haven't reproduced it in a test environment yet.

rofrischmann:
Oh sure, I didn't think about updating the monolithic renderer at all. Will try to look into that, but I am not using (and would never recommend to do either)

vlad-zhukov:
I find it handy during development and testing.

rofrischmann:
Alright, looks like you're doing it on the client right? Can't reproduce that locally. Can you try it with <Provider rehydrate={false} />?

vlad-zhukov:
I'll have a look tomorrow.

rofrischmann:
Great! Maybe we should switch the rehydration default to false to one can enable it for prod-only (atomic design). Its not compatible with monolithic rendering

TxHawks:
I'd keep the default as is, since atomic design is also the the default in development mode. There should just be an explanation making clear how to switch it off in the `fela-monolithic` readme.
Or do you mean turning it off only when `fela-monolithic` is actually used?

rofrischmann:
I thought about it again and think the renderer should be the single source of truth. Therefore I added a "feature flag" which I can simply turn off if monolithic is applied.
This could be a general approach for different features!

vlad-zhukov:
Not sure what you mean, but can you open this issue back please? I think I've found how to reproduce it in an isolated environment.

rofrischmann:
Got that fixed as you can see on Travis :) will publish tomorrow

vlad-zhukov:
Thanks! Is it going to fix a similar bug with atomic classes?

PR #418, opened 2017-10-02
[WIP] Add a failing test for fela-monolithic usage with React

vlad-zhukov:
This is a repro of #395

### DIFF
--- a/packages/fela-monolithic/src/__tests__/__snapshots__/integration-react-test.js.snap
+++ b/packages/fela-monolithic/src/__tests__/__snapshots__/integration-react-test.js.snap
@@ -1,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Monolithic enhancer React integration should render a component 1`] = `".f137u7ef{color:red}.f137u7ef{color:red}"`;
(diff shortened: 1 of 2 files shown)


packages/fela-monolithic/src/__tests__/__snapshots__/integration-react-test.js.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Monolithic enhancer React integration should render a component 1`] = `".f137u7ef{color:red}.f137u7ef{color:red}"`;
```
vlad-zhukov: There should be a single class like it is in v5.

vlad-zhukov:
It shouldn't be merged 😄

rofrischmann:
Hum why not :P it's a valid test case I guess

vlad-zhukov:
Because right now it outputs `.f137u7ef{color:red}.f137u7ef{color:red}` instead of `.f137u7ef{color:red}`. The CI is failing due to lerna bootstrap error. I should've also be using equality test instead of snapshotting.

rofrischmann:
snapshot is totally fine, we just need to fix it first, but I already got that working I guess

PR #422, opened 2017-10-05
Update withTheme docs

aga5tya:
Minor updates to documentation.


packages/react-fela/docs/createComponent.md
```diff
@@ -42,12 +42,14 @@ Using the `passThroughProps` parameter allows us to pass props to the underlying
 
 If passing a className, it will automatically be concatenated with the Fela generated className. This allows composing multiple Fela Components.
 
-#### Functional passThroughProps
+#### Functional passThroughProps [Deprecated: Read Note Below]
```
rofrischmann: First of all, thanks for the update!

Would you please remove the Deprecation sign? It is not deprecated and withProxy is still not recommended as it doesn't work with nested props. Just keep the basic NOTE: and its fine :) (btw. switch "NOTE:" to "> Note:" if possible, thats how we did everywhere by now.

packages/react-fela/docs/withTheme.md
```diff
@@ -25,3 +28,8 @@ const RenderTreeFragment = (
 
 // => <div>The primary color is red</div>
 ```
+## Note
```
rofrischmann: This could be renamed to Tips as we did that at some other places as well :)

aga5tya: @rofrischmann changes done. Thanks for pointing them out 👍

PR #441, opened 2017-10-19
WIP: Yarn workspace

tajo:
- example packages moved into `/examples` and renamed
- lerna updated
- yarn workspaces (root `yarn install` rules them all)

Weird issues:

```
yarn install
cd examples/react
yarn start 
```

throws

```
Error: Cannot find module 'react-fela'
```

Which is really weird because webpack should resolve it from the root `node_modules`. Not sure why that happens. :/ Also for other examples, it doesn't add `.bin` symlinks. Maybe a bug in yarn workspace?

todo

- add conventional commits
- update docs
- delete `bootstrap` command, it's not necessary


lerna.json
```diff
@@ -1,9 +1,11 @@
 {
-  "lerna": "2.0.0-rc.4",
+  "lerna": "2.4.0",
   "version": "independent",
   "npmClient": "yarn",
+  "useWorkspaces": true,
   "packages": [
```
vlad-zhukov: These are not needed anymore

vlad-zhukov:
> webpack should resolve it from the root `node_modules`

Webpack has never done that, the root is the folder you run it in, i.e. `process.cwd()`.

tajo:
@vlad-zhukov 

https://webpack.js.org/configuration/resolve/#resolve-modules says

> A relative path will be scanned similarly to how Node scans for node_modules, by looking through the current directory as well as it's ancestors (i.e. ./node_modules, ../node_modules, and on).

Afaik, it should go through all ancestor directories.

vlad-zhukov:
But it has to stop looking at some point, right?

> similarly to how Node scans for `node_modules`
Node only resolves modules inside the current working directory.

You will need to add `modules: [path.resolve(__dirname, '../..'), 'node_modules']` to webpack configs to make it work.

rofrischmann:
What's the status on this one? I'm not really into the internals, so I can't tell if I can merge it already?

tajo:
@rofrischmann still in WIP, will look into it early next week.

TxHawks:
It's important to keep in mind that there is currently a bug with running lifecycle methods when linking local packages.
See [here](https://github.com/yarnpkg/yarn/issues/3911) for more details

rofrischmann:
Good to know @TxHawks.
We don't use them internally, but maybe some deps?

rofrischmann:
Can I somehow finish this one? Would like to merge asap as I might take some time off and wanna clean up before :) @tajo

rofrischmann:
@tajo I'll close this for now if that's fine with you.
Would love to see it done, but I'd also like to clean-up the repo.

PR #447, opened 2017-10-23
Make preset-web configurable

AugustinLF:
Makes it possible to pass options to `fela-preset-web`. I made `createWebPreset` ignorant of the options of each plugins (hence spread and the default array), if you'd prefer me to make it explicit, it's doable.

#### Patch
- fela-preset-web
- fela-preset-dev

### DIFF
--- a/packages/fela-preset-dev/README.md
+++ b/packages/fela-preset-dev/README.md
@@ -31,6 +31,25 @@ const renderer = createRenderer({
 })
 ```
 
+You can also pass options to the plugins:
+```javascript
+import { createRenderer } from 'fela'
+import {createDevPreset} from 'fela-preset-dev'
+
+const renderer = createRenderer({
+  plugins: [
+    ...createDevPreset({
+      'fela-plugin-validator': [
+        {
+          logInvalid: true,
+          deleteInvalid: true
+        }
+      ]
+    })
+  ]
+})
+```
+
 ## License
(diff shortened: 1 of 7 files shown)


rofrischmann:
Hmm. Isn't this kind of controversy? I thought you'd use presets especially because you don't have to configure anything :P 
You could just configure your renderer manually if you need different stuff?

AugustinLF:
I understand, you do have a point, it's just that when doing upgrades, it's usually easier to work with presets than with several plugins^^

You can also see preset as a base to work with, something to remove the burden of choice for the user. This makes on boarding easier!

If you look at how babel works, the `babel-preset-env` is a bundle of different transforms, but you can still configure it if you want!

rofrischmann:
Alright alright, I get the idea and it's reasonable.
Would you also apply this to the dev preset? Just for consistency.

AugustinLF:
@rofrischmann Done!

rofrischmann:
Great!
It's backwards compatible, isn't it?

AugustinLF:
It should be. I added the tests for the default case, and it seems to work as expected. I never used `fela-preset-dev`, so on this one I'm less confident, but I don't see why it shouldn't work. I'll retest the preset-web against the my work project later this afternoon, just to be sure :)

AugustinLF:
Tested again, it works perfectly!

PR #471, opened 2017-11-15
[TypeScript] Fix combineRules() typings

### LINKED ISSUE #377: [TypeScript] 'combineRules' ignores rule props
**Version**: 5.2.0
**Environment**: DOM / Server / Both
**Type**: Bug

-------------------
#### Description

The current definition for combineRules ignores rule parameter types and therefore the compiler will miss a lot of potential errors

#### Steps to reproduce

1. Create rules with props:

```typescript
import { createRenderer, combineRules } from 'fela'

const rule = (props: { fontSize: number }) => ({
  fontSize: props.fontSize,
  color: 'red'
})

const anotherRule = () => ({
  color: 'blue'
})

const superRule = combineRules(rule, anotherRule)
```

2. Render the combined rule with an incorrect parameter

```typescript
renderer.renderRule(superRule, { fontSize: '12px', foo: 'bar' })
```

#### Expected Behavior

I'd expect the definitions to be defined with generics such that you could call the following and get the compiler error

```typescript
renderer.renderRule(superRule, { fontSize: '12px', foo: 'bar' })
                                                   ^^^^^^^^^^
                                              is not a property of the params
```

#### Actual Behavior

The actual behavior is no warning at all. This manifests itself in react-fela as components being passed props that aren't defined on the component (and thus a compiler error).

jrf0110:
Closes #377 

__Error 1 caught:__

image

__Error 2 caught:__

image

__All good__

image

### DIFF
--- a/packages/fela/index.d.ts
+++ b/packages/fela/index.d.ts
@@ -3,7 +3,7 @@ declare module "fela" {
   import { CSSProperties } from 'react';
 
   type TRuleProps = {};
-  type TRule = (props: TRuleProps) => IStyle;
+  type TRule<T = TRuleProps> = (props: T) => IStyle
   type TKeyFrame = TRule;
   type TRendererCreator = (config?: IConfig) => IRenderer;
   type TPlugin = (style: IStyle) => IStyle; //http://fela.js.org/docs/advanced/Plugins.html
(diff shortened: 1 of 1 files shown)


rofrischmann:
Amazing. Thank you!

PR #478, opened 2017-12-01
Feature/bug animation important

adderpositive:
## Description
Adds `!important` to every declaration value except animated items (not by `transition`) and keyframes. Animation items and keyframes with `!important` disrupt functionality of animations. [More about.](https://developer.mozilla.org/en-US/docs/Web/CSS/%40keyframes#!important_in_a_keyframe) 

For every items which includes any property of `animation*:` should be set **manually**, because not possible to catch, what property will be changed by `keyframes`. 
Every `keyframes` are catched according to key property `to`, `from` and if they contains `%`.

### DIFF
--- a/packages/fela-plugin-important/README.md
+++ b/packages/fela-plugin-important/README.md
@@ -2,7 +2,7 @@
 
 <img alt="npm version" src="https://badge.fury.io/js/fela-plugin-important.svg"> <img alt="npm downloads" src="https://img.shields.io/npm/dm/fela-plugin-important.svg">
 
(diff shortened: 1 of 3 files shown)


rofrischmann:
Oh interesting! I will review the changes as soon as possible. Thank you for your PR 🤗

rofrischmann:
Looks fine to me! Thank you =)

adderpositive:
No problem. My pleasure. :) Yeah, output will be changed a little bit.

PR #481, opened 2017-12-05
Doc fix: 'named-media-query' should be 'named-keys'

shanewholloway:
`fela-plugin-named-media-query` was renamed to `fela-plugin-named-keys` in commit 1b2257d053568745f58fa0e8889f9f0120e2795e


rofrischmann:
Thank you!

Issue #482, opened 2017-12-05
Extending styles implementation for connected components


sgrishchenko:
**Type**: Feature

| Packages | Version |
| -------- | ------- |
| fela-tools     | 5.0.x   |
| fela-bindings     | 1.2.x   |

-------------------
#### Description
I propose to implement a mechanism for extending styles for components created using the `connect` functions.

#### Proposal
It might look like this:

**Code Example**:
```javascript
import { connect } from 'react-fela'

const rules = props => ({
  rule1: {
    padding: 1,
  },
  rule2: {
    color: props.color,
  },
})

const MyComponent = connect(rules)(({ styles }) => (
  <div>
    <span className={styles.rule1} />
    <span className={styles.rule2} />
  </div>
))

const extend = {
  rule1: {
    padding: 2
  },
  rule2: {
    fontSize: 16
  }
}

<MyComponent extend={extend} />
// => <div>
// =>  <span className="a" />
// =>  <span className="b c" />
// => </div>
// => .a { padding: 2 }
// => .b { color: red }
// => .c { font-size: 16 }
```

PR #483, opened 2017-12-05
Extending styles implementation for connected components.

sgrishchenko:
## Description
see https://github.com/rofrischmann/fela/issues/482

## Example
see https://github.com/sgrishchenko/fela/blob/8f92affadfbcc1b8e03a0e29242500a04497c4e9/packages/react-fela/docs/connect.md

#### Minor
* fela-tools
* fela-bindings
* react-fela
* preact-fela
* inferno-fela

## Checklist

#### Changes


rofrischmann:
Hmm. While I really like the idea, this looks far too complex and edgy to be implemented in Fela's core (e.g. combineMultiRules is rather a helper than a core API).

I have to think about it again, but wouldn't it make more sense to create a custom connect package being sth. like "Advanced connect-API for Fela"?

I'd rather want to keep the core simple at this point.
Yet, I would love to help to extract that into a separate package!

sgrishchenko:
For me, it does not matter where the combMultiRules will be. I placed this function there only because on functionality it's very similar to combineRules. I can put it in a separate package or put it into an existing one, for example, in fela-utils or fela-tools.

sgrishchenko:
Yes, it seems to me fela-tools is a great place. Why not?

rofrischmann:
Yeah fela-tools is made especially for advanced APIs like that. would be great to place that connect extension to fela-tools and I'll happily merge it =)

sgrishchenko:
Ready =)

TxHawks:
Don't the docs need updating now?

sgrishchenko:
I think I described it pretty well [here](https://github.com/sgrishchenko/fela/blob/8f92affadfbcc1b8e03a0e29242500a04497c4e9/packages/react-fela/docs/connect.md#extending-styles). The last changes do not concern functionality, they concern only the location of the utility function. The API remains the same.

sgrishchenko:
If you have any questions about the API you can ask them and I will certainly answer the documentation ;-)

packages/react-fela/docs/connect.md
```diff
@@ -79,3 +79,100 @@ ReactDOM.render(
...
+})
+
+const MyComponent = connect(rules)(({ styles }) => (
+  <div>
+    <span className={styles.rule1} />
+    <span className={styles.rule2} />
+  </div>
+))
+
+const anotherRules = {
```
TxHawks: I'd change `anotherRules` to `extraRules` or `moreRules`, just for clarity.

packages/react-fela/docs/connect.md
```diff
@@ -79,3 +79,100 @@ ReactDOM.render(
...
+  rule1: () => ({
+    padding: 2
+  }),
+  rule2: () => ({
+    fontSize: 16
+  })
+}
+
+const ProxyWrapper = connect(anotherRules)(({ rules: injectedRules }) => (
```
TxHawks: Where is `injectedRules` coming from?

packages/react-fela/docs/connect.md
```diff
@@ -79,3 +79,100 @@ ReactDOM.render(
...
+))
+```
+
+Inside the component, you get the `rules` always in a normalized state.
+This means that the `rules` will have roughly the following form:
+
+```javascript
+const rules = (props) => ({
+  rule1: (props) => ({/* some props adn values */}),
+  rule2: (props) => ({/* some props adn values */}),
```
TxHawks: `adn` should be `and` (here and in the line above)

packages/react-fela/docs/connect.md
```diff
@@ -79,3 +79,100 @@ ReactDOM.render(
...
+
+const ProxyWrapper = connect(anotherRules)(({ rules: injectedRules }) => (
+  <div>
+    <MyComponent color="red" extend={injectedRules} />
+  </div>
+))
+```
+
+Inside the component, you get the `rules` always in a normalized state.
+This means that the `rules` will have roughly the following form:
```
TxHawks: For clarity, I'd change the above two lines to:
```
The underlying component is eventually passed a normalized `rules` object:
```

I think the brevity makes it easier to understand

packages/react-fela/docs/connect.md
sgrishchenko: This variable is named an "anotherRules"  because these rules do not extend and do not complement something. These are the rules for the ProxyWrapper. Maybe I should call them componentRules and proxyWrapperRules?

packages/react-fela/docs/connect.md
sgrishchenko: The variable injectedRules is the resulting multi-rule, on the basis of which the property "styles" was formed. This property is injected so that the parent component has the ability to control the styling of the child components on the basis of its own styles. This should be described in the documentation?

packages/react-fela/docs/connect.md
sgrishchenko: Thank you, I will correct this.

packages/react-fela/docs/connect.md
sgrishchenko: Yes, I will correct it.

packages/react-fela/docs/connect.md
TxHawks: Yes, I'd opt for explicitness whenever possible. It decreases the cognitive load significantly and makes docs easier to understand

packages/react-fela/docs/connect.md
TxHawks: That's what I meant. Add a comment above to explain what's going on there. Documentation should be just as succinct as possible, but very clear to an inexperienced user.
Showing use examples is great and super helpful, but above all, I think good documentation should aim to have people easily understand what's going on

TxHawks:
@sgrishchenko - just wanted to say that I hope my comments don't discourage you.

I think you did a great job on this, and am just nitpicking because I feel good and clear documentation is crucial, especially for more advanced techniques.

Thank you for the great work and your patience

sgrishchenko:
Я полностью согласен, что документация очень важна. Я попрытался внести описанные выже предложения. Могу ли я попросить посмотреть на документацию снова?

PR #495, opened 2018-01-08
[PoC] Deterministic Class Names

rofrischmann:
## Description
Adds experimental support for deterministic class names.
This PoC is made to show how it could look like, it makes a lot of internal calculations more simple, stable and straight-forward. Yet it increases the overall bundle/markup size as class names are now ~6 chars long. The calculation generates an unique hash (using string-hash) which is then converted to a Base36 string.

## Example

```javascript
const rule = () => ({
  color: 'red',
  backgroundColor: 'blue'
})

const anotherRule = () => ({
  color: 'red',
  backgroundColor: 'green'
})

renderer.renderRule(rule)
// => _gshHad _he3asi

renderer.renderRule(anotherRule)
// => _h3pyw _he3asi
```

#### Major
* fela

#### Patch
* fela-dom

## Checklist

#### Changes


tajo:
Interesting. Is it going to be opt-in type of settings? When do you use this?

rofrischmann:
I considered making this the default while the opinionated counter-based solutions is turned into an enhancer as this is more bulletproof and safe to use for beginners and makes rehydration much simpler as well.

sgrishchenko:
I see the use of a hash function (string-hash) but I do not see the collision resolution algorithm. Is it possible that the same CSS class name will be generated for different CSS rules? And if not, why not?

rofrischmann:
I do not have any tests for that, but from what I've read this is one of the most safe hashing algorithms in terms of collision and in all tests I did not see any colliding class name either.
If you see any way to improve that, feel free to propose! I am not really into hashing algorithms at all.

sgrishchenko:
This test shows that there are now problems: 

`it('should not cause a collision between classNames', () => {
      const rule1 = () => ({
        ':after': {
          content: 'go'
        }
      })

      const rule2 = () => ({
        ':after': {
          content: 'Fn'
        }
      })

      expect(renderer.renderRule(rule1)).not.toBe(renderer.renderRule(rule2))
    })
  })`

sgrishchenko:
now this test failed

rofrischmann:
I might just not include that PR then :) 
Wasn't aware that so simple cases already fail so it's more than great to have collision detection here. Yet, I don't think using a single cache is a great idea.

If you got 1 renderer that has a selectorPrefix "foo-" and the other is default. Now you render color:red with the first one and get foo-xyz. Now you render that with the default one again and you'll get the foo-xyz class name as well, but actually don't want that prefix there.

(That's kind of the only reason why you would have multiple caches anyway)

sgrishchenko:
See please https://github.com/rofrischmann/fela/pull/506. I hope I solveв the problem with collisions.

rofrischmann:
Closing as it causes collisions that again cause issues with server-rendered and client-rendered indifferences.

PR #500, opened 2018-01-18
Render-props APIs

rofrischmann:
## Description
Adds two new APIs using the render-prop pattern. This has many benefits over the HoC approach.

## Example
```javascript
<FelaComponent
  rule={...} 
  render={({ className, children }) => (
    <div className={className}>{children}</div>
  )
/>

<FelaTheme 
  render={theme => (
    <div>The theme is {JSON.stringify(theme)}</div>
  )
/>
```

#### Minor
- fela-bindings
- react-fela
- inferno-fela
- preact-fela

## Checklist

#### Changes


TxHawks:
👍  * 💯 !

If I understand correctly, `<FelaTheme>` is the equivalent of `withTheme`, correct? I may be wrong and there actually is a use case, but I don't think it's actually needed with the render prop approach, where `theme` is available in the scope of the underlying component.

rofrischmann:
If youre not using a render prop for that as well, you have to got to static hoisting etc. again which kinda sucks. I feel like the render prop approach is way cleaner and simpler (and probably faster and smaller in bundle size)

TxHawks:
Im not sure I understand...

Will this not work?

```jsx
<FelaComponent 
  rule={style}
  render={({theme, ...props}) => (
    <div>The small breakpoint is {theme.bps.small}</div>
  )
} />
```

rofrischmann:
Yeah, it does. But still, you need the `FelaTheme` render-prop API to compose the `FelaComponent` with it. Using `withTheme` there would simply deny all the benefits.

TxHawks:
Not sure I follow. Can you show a coffee example?

rofrischmann:
Well, you can see, that `FelaComponent` uses `FelaTheme` internally (https://github.com/rofrischmann/fela/blob/7236c25aaee2cb225aee4ee179b601b8428bb457/packages/fela-bindings/src/FelaComponentFactory.js#L9)

This could be achieved using the `withTheme` as well, but then again you would have a HoC and would need to hoist statics and combine contextTypes and whatever which seems not optimal.

Also I believe if we provide some render-prop stuff for `FelaComponent` it would be more than fair to also have `FelaTheme`, because I can think of many use cases where you'd need the theme without actually styling a component.

rofrischmann:
Yeah of course, once everything is done and tested I will add official documentation :)

Other question:
What's yet to be solved is how composition would work. Feels like this is something that must be handled manually which is something I don't really like actually:

```javascript
const button = () => ...

const Button = ({ rule }) => (
  <FelaComponent 
    rule={combineRules(button, rule)} 
    render={({ className }) => <div className={className}>Click me</div>}
  />
)

const blueButton = () => ...

const BlueButton = () => (
  <Button rule={blueButton} />
)
```

Or would you just go and reuse the button rule above directly?
Could also have a extend prop that automatically combines rules, but it'd still need to be passed down to the FelaComponent:

```javascript
const Button = ({ extend }) => (
  <FelaComponent 
    rule={button}
    extend={extend}
    render={({ className }) => <div className={className}>Click me</div>}
  />
)
```

TxHawks:
I actually like the use of `combineRules()` fire composition as it is very explicit, but still think an `extend` prop is a good idea

rofrischmann:
Closing in favor of the release branch.

PR #501, opened 2018-01-19
Babel Plugin Hashing Mechanism

rofrischmann:
closes #497

## Description
Adds a simple `toString(36)` in order to turn the number hash into a short string starting with a letter. This way they can be accessed via direct object property accessors without yielding errors.

#### Patch
* babel-plugin-fela

## Checklist

- [x] The code was formatted using Prettier (`yarn run format`)
- [x] The code has no linting errors (`yarn run lint`)
- [x] All tests are passing (`yarn run test`) 
- [x] There are no flow-type errors (`yarn run flow`)

#### Changes
If one of the following checks doesn't make sense due to the type of PR, just check them.

- [x] Tests have been added/updated
- [x] Documentation has been added/updated
- [x] My changes have proper flow-types

PR #506, opened 2018-01-24
Prevent collisions in hash generation

sgrishchenko:
## Description
Prevent collisions in hash generation

#### Patch
fela

## Checklist

#### Changes

### DIFF
--- a/packages/fela/src/__tests__/createRenderer-test.js
+++ b/packages/fela/src/__tests__/createRenderer-test.js
@@ -209,6 +209,24 @@ describe('Renderer', () => {
 
       expect(renderToString(renderer)).toMatchSnapshot()
     })
+
+    it('should not cause a collision between classNames', () => {
+      const rule1 = () => ({
+        ':after': {
+          content: '"go"',
+        },
+      })
+
+      const rule2 = () => ({
+        ':after': {
+          content: '"Fn"',
+        },
+      })
+
+      const renderer = createRenderer()
+
+      expect(renderer.renderRule(rule1)).not.toBe(renderer.renderRule(rule2))
+    })
   })
 
   describe('Rendering keyframes', () => {
--- a/packages/fela/src/generateUniqueHash.js
+++ b/packages/fela/src/generateUniqueHash.js
@@ -1,6 +1,18 @@
 /* @flow  */
 import hash from 'string-hash'
 
+const collisionBuffer = {}
+
 export default function generateUniqueHash(reference: string): string {
-  return '_' + hash(reference).toString(36)
+  let result = reference
+  let prevResult
+
+  do {
+    prevResult = result
+    result = `_${hash(prevResult).toString(36)}`
+  } while (collisionBuffer[result] && collisionBuffer[result] !== prevResult)
+
+  collisionBuffer[result] = prevResult
+
+  return result
 }


rofrischmann:
What happens if I have two renderer with different selectorPrefix set?

```javascript
const rule = () => ({ color: 'red' })

const renderer2 = createRenderer({ selectorPrefix: 'foo-' })

// does this still work?
renderer.renderRule(rule) // => _abc
renderer2.renderRule(rule) // => foo-_abc
```

sgrishchenko:
Yes, I don't change _createRenderer_ realization. The trick in that ```collisionBuffer``` is singleton. It means that ```generateUniqueHash``` function will work, until return **realy unique** hash code for inputed key. I always keep pair of key and value, where key is hash and value is key for what this hash was generated. And if in ```collisionBuffer``` yet exist key with same hash but value for this hash is not equal inputed reference, it means that collision existing and I need generate hash again. This is a common method of solving this problem.

sgrishchenko:
More interested next situation. For example you have 2 rules, that raise collision (**rule1** and **rule2**). And you have next code:
```js
const rule1 = () => ({ color: 'red' })
const rule2 = () => ({ color: 'blue' })

const renderer1 = createRenderer()
const renderer2 = createRenderer({ selectorPrefix: 'foo' })

renderer1.renderRule(rule1) // => _abc
renderer2.renderRule(rule1) // => foo_abc
renderer1.renderRule(rule2) // => _123
renderer2.renderRule(rule2) // => foo_123
```  
but in another environment (for example on another tab in brauser) you change order `renderRule` calls, you get
```js
renderer1.renderRule(rule2) // => _abc
renderer2.renderRule(rule2) // => foo_abc
renderer1.renderRule(rule1) // => _123
renderer2.renderRule(rule1) // => foo_123
```
**Note** that I swapped the rules, but the order of the classes did not change. Is this behavior right for you?

rofrischmann:
Hmm. That's the whole reason why I introduced hash class names: to make it predictable and unique. How can the processed hash of `color` and `red` be the same as `color` `blue`? Or is it just a problem with this example. This must not happen or we can just stick to the counter method anyways.

TxHawks:
How does this and #501 effect Fela's performance beyond bundle size and html size? Generating class names is definitely a hot path in Fela. I can't say I know, but I'd imagine hashing is much more expensive than counter-based opperations

rofrischmann:
It definitely is slower as there's no faster way to generate an unique id using a counter, but it sure comes with a price. It makes Fela much more explicit and simple to use and you don't have filterClassName (and most likely don't need selectorPrefix) at all.

The perf impact can be measured with benchmarks, but is not really helpful to test real world applications at all, because both are extremely fast and with most SPAs you got >80% (more likely to be >90%) cache hits anyways.

Yet, if you're a pure perf junkie or simply need to be as fast as anyhow possible, you will be able to use an enhancer that adds the old behaviour anyways (not yet implemented, but will be added before release to keep backwards compatibility high).

sgrishchenko:
It just a problem with this example. With resolving collisions you have no another way. You don't know about collisions until its happens. But on single environment you always have same order of collision resolutions. But I see potential problem. For example you have server side rendering. On server you may have another sequence rule rendering. And then you try render rule on client, you again get collisions.

rofrischmann:
Hmm. So since we need collision resolution to make it 100% safe, we again bound to some context (order / buffer) which makes in unpredictable in general once again. So my hope for this method is kinda not possible here. Without 100% predictability, I would rather go and use the counter method instead, as it has the same problems, but is at least collision free in context.

sgrishchenko:
Why we can't use compression instead hashing?

sgrishchenko:
Or use references as class names?

rofrischmann:
How would you compress the references? reference as class names would need to be escaped and can be enormously long which would drastically bloat the HTML and CSS output for SSR and would not really read well. Any way to make the className (reference) smaller and yet unique is highly welcome.

sgrishchenko:
I tried to study approaches to compression and choose the appropriate one for this case. Compression based on the use of utf8 is poorly suited because there are problems with escapin. The only thing that's less suited to this is Base64 compression. But the benefit from it is only visible if the string is longer than 25 characters. How do you think this method is like?
I find usefull util library for comperrion https://www.npmjs.com/package/lz-string

rofrischmann:
Hows the overal performance impact with those? I'm wondering if it's really worth it. Right now, I am about to revert the PR and use the counter method again.
25+ chars is a rare find I guess (maybe media + pseudo or sth. like that) and in general I'd avoid class names > 5-6 chars as otherwise the atomic approach would bloat the HTML too much.

sgrishchenko:
If you want to have class names no longer than 5-6 characters, standard text compression methods will not work. Theoretically, this effect can be achieved by writing your own compressor, which will know that we are compressing CSS. For example, keywords will be coded with a single character. But what to do with places where an arbitrary text can be contained? (e.g., content) In any case, in such conditions it becomes not a trivial task. The most annoying thing is that I just realized that server rendering is not available for Fela right now ...

rofrischmann:
? Server rendering is pretty straightforward http://fela.js.org/docs/advanced/ServerRendering.html

sgrishchenko:
Oo... I'm sorry. I forgot about rehydrate...

PR #510, opened 2018-02-04
[fela-plugin-custom-property] Avoid deleting resolved custom properties

HugoGiraudel:
## Description

This pull-request fixes a minor bug with the `fela-plugin-custom-property` plugin.

When naming a custom property after one of the CSS properties it eventually outputs, that very property gets deleted. It works well in the case of the `size` example as `size` should not be present in the output object, but when using this plugin a custom enhancer or validator, it doesn’t yield expected results.

## Example

```javascript
const shorthand = property => value => {
  console.warn(`The \`${property}\` should not be used because of lack of reusability of shorthand properties in Fela. Use the sub-properties instead.`)

  return { [property]: value }
}

export default createRenderer({
  plugins: [ customProperty({ padding: shorthand('padding') }) ],
  enhancers: [ … ]
})

```

## Versioning

`fela-plugin-custom-property` should be bumped to 6.0.4.

## Checklist

#### Changes

### DIFF
--- a/packages/fela-plugin-custom-property/src/__tests__/customProperty-test.js
+++ b/packages/fela-plugin-custom-property/src/__tests__/customProperty-test.js
@@ -56,4 +56,18 @@ describe('Custom property plugin', () => {
       },
     })
   })
+
+  it('should not remove resolved properties', () => {
+    const padding = value => ({ padding: value })
+
+    const style = {
+      padding: '1em',
+    }
+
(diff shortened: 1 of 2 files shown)


rofrischmann:
Yey, thank you for fixing this :)

HugoGiraudel:
I hope I didn’t break anything. ;)

PR #522, opened 2018-02-28
don't resolve nested style objects if property was removed

oleksiibilous:
## Description
If value is object it should be resolved as nested style object only if its property was not removed before. Righ now I get warnings each time I use custom propertiy which receives object 

## Example

```javascript
export default createRenderer({
  plugins: [
    customProperty({
      padding: ({ t, l, b, r }) => {
        const obj = {}

        if (t != null) obj.paddingTop = t
        if (l != null) obj.paddingLeft = l
        if (b != null) obj.paddingBottom = b
        if (r != null) obj.paddingRight = r

        return obj
      }
    })
  ],
})
```

#### Patch

- fela-plugin-custom-property

## Checklist

#### Changes

### DIFF
--- a/packages/fela-plugin-custom-property/src/__tests__/customProperty-test.js
+++ b/packages/fela-plugin-custom-property/src/__tests__/customProperty-test.js
@@ -70,4 +70,21 @@ describe('Custom property plugin', () => {
       padding: '1em',
     })
   })
+
+  it('should not resolve nested style objects if property was removed', () => {
+    expect(
+      customProperty({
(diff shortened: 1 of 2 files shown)


rofrischmann:
Oh true, thanks for adding that!

oleksiibilous:
Sure, my fault😁 I wanted to mark that as a patch but probably missed the line. Fixed description, thnx

menelike:
fixes https://github.com/rofrischmann/fela/issues/524

PR #523, opened 2018-03-04
Add Net-A-Porter to list of companies using Fela

lucasmotta:
## Description
We have been using Fela on our latest projects at Net-A-Porter. It's a relatively large application, so we are pretty happy with the results so far!

#### Changes


rofrischmann:
Yey! Cool, thank you :)

PR #553, opened 2018-04-13
ThemeProvider code example was missing a character

aaronschubert0:
Thought I'd submit a fix while I was reading

#### Changes


TxHawks:
Thanks!

Issue #552, opened 2018-04-11
fela-plugin-validator to support CSSLint rule configuration or root .csslintrc file


brianespinosa:
#### Proposal
Currently when CSSLint is enabled, I cannot see a way for us to pass a configuration to it so we are forced to use all of the default rules. You can technically disable rules with the per-line syntax as long as you include that comment as part of your style definition.

The problem with this is the comment then gets rendered into the atomic class/style.

**Current Example**:
```javascript
// CURRENT:
const rule = (props) => {
  const {theme} = props;

  return {color: theme.primaryColor + ' !important /* csslint allow: important */'};
};
```
It would be great if we could simply look in the root folder for the .csslintrc file and pass that as a parameter in the plugin automatically. Or at least be able to pass that as a config option to fela-plugin-validator.

I guess the per-line option is good if we were stripping out CSS comments in Fela. I am not aware of an option for this.

sgrishchenko:
I have not ideas how support .csslintrc file because linting performed in browser and for this to work you need pack .csslintrc into bundle. But I will support passing js object into plugin configs.

PR #554, opened 2018-04-14
fela-plugin-validator to support CSSLint rule configuration

sgrishchenko:
## Description
https://github.com/rofrischmann/fela/issues/552

## Example
Using validator plugin with CSSLint rule configuration

```javascript
validator({
  useCSSLint: {
    important: false,
  },
})
```

#### Minor
fela-plugin-validator

## Checklist

#### Changes


rofrischmann:
Is it backwards compatible with `useCSSLint: true`?
Otherwise we need a major update :P 

Anyways, will check that PR in the afternoon (y)

sgrishchenko:
Yes, it backwards compatible. If you set `useCSSLint: true` you will use config by default, as before. I did not correct previous tests.

PR #556, opened 2018-04-19
Missing dependency for "example-inferno"

meydjer:
"inferno-create-element" was missing for "example-inferno".


rofrischmann:
Thanks a lot!

PR #559, opened 2018-04-26
Fix rehydrated twice generated monolitihic classes

tomwagner:
## Description
This is related with old closed issue #395. You need to use monolithic enhancer. It causes twice rendered classes in DOM (is annoying for development). The first set is from SSR and next set is generated because of wrongly cached classes to monolithic usage. Fela tries to rehydrate generated monolithic classes. They are rehydrated properly, but not for monolithic usage. There are too long keys in renderer cache (because of multiple rules per class). So i added repair function to fix rehydrated cache to work properly with monolithic enhancer.

## Example
```javascript
  plugins: [
    ...
  ],
  enhancers: [monolithic()]
});
```

#### Patch
fela-dom
fela-monolithic

## Checklist

#### Changes

### DIFF
--- a/packages/fela-dom/src/dom/rehydration/rehydrate.js
+++ b/packages/fela-dom/src/dom/rehydration/rehydrate.js
@@ -34,4 +34,6 @@ export default function rehydrate(renderer: DOMRenderer): void {
       }
     }
   })
+
+  renderer._emitChange({ type: 'REHYDRATATION_FINISHED' })
 }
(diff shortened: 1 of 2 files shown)


packages/fela-dom/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fela-dom",
-  "version": "7.0.7",
+  "version": "7.0.8",
```
TxHawks: No need to manually bump the version, the release process will take care of it

packages/fela-monolithic/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "fela-monolithic",
-  "version": "5.0.21",
+  "version": "5.0.22",
```
TxHawks: Same as above

packages/fela-dom/package.json
tomwagner: Removed, thanks for notify.

PR #564, opened 2018-05-02
Fe

rofrischmann:
## Description
Fe is for Fela, what glam is for glamor:
A `createElement` replacement which is used to compile JSX.

## Example

```javascript
/* @jsx fe */
import { fe } from 'react-fela'

const style = {
  color: 'red',
  ':hover': {
    color: 'blue'
  }
}

ReactDOM.render(
  <div style={style}>
    Hello
  </div>
)

// => <div class="a b">Hello</div>
```

#### Minor
* fela-bindings
* react-fela
* inferno-fela
* preact-fela

## Checklist

#### Changes


TxHawks:
This looks great.

Some thoughts on the API:
1) Maybe it's worth considering using `css` as the API prop instead of `style`, that way we don't clobber any native APIs, and both behaviors are available to the user.
2) Since this is already using `FelaComponent` under the hood, we could offer the same capabilities of its `rule` prop at no extra cost. We just use `FelaComponent`'s `style` method if the `style` prop is passed an `object`, and its `rule` method when it's a function. I think having access to the `theme` and extra props can prove very useful.
3) Since `FelaComponent` uses `customClass` for concatenating additional classes to the generated ones, I think we should use it here as well. I think it is important to keep the API consistent across the lib as much as possible, but also, it is more explicit (this is true for both `style` and `className`. When I read 
```jsx
<div className="some-extra-class" style={{ color: 'red', }}>I'm red</div>
```

It is not immediately clear that this will not behave the same way ordinary JSX does. However, reading the following, it is immediately clear that there something going on beyond vanilla JSX 
```jsx
<div customClass="some-extra-class" css={{ color: 'red', }}>I'm red</div>
```

rofrischmann:
Yeah, had similar ideas and questions lately.

1. Also considered the `css` prop, just like glam does it, but maybe I want an option (maybe just another import) that directly uses `style`, so it can be used as a drop-in replacement for inline style based applications. The reason for this is simple: I'm right now building a huge app and prototyped everything with just `<div>` and inline style, which I'd like to replace with Fela with no cost later on ;)

2. Yeah true, but I'm not sure if you'd actually use functions here, except for accessing `theme` which is already possible with `style` as well `const style = theme => ({ })`.

3. That's something I'm not agreeing actually. It is meant to be a JSX replacement, so we should use default APIs (at least for 'className'). Otherwise you'd not be able to use it as a drop-in for any inline style apps. Only thing I'm okay with is the `css` prop, as this stills allows the use of inline style, which sometimes is pretty useful when overriding third-party styles.

TxHawks:
I think a codmod is probably the best option for handling the `style` <-> `css` prop conversion use case.

As for using functions, It may not be the most frequent use case, but I can definitely imagine some. Here is one:

```jsx
const style = ({ typescaleStep, theme, }) => ({
  ...setTypescale(typescaleStep)  // retruns an object with font-size and line-height
                                  // based on a predefined typographic scale
})

<div typescaleStep={2} style={style}>Text that's a bit bigger than default</div>
```

As for number 3, fair enough.

rofrischmann:
But why wouldn't you just pass the `typescaleStep` within the style directly? style vs. css looks fine to me, let's change that.

Having the `rule` additionally looks fine to me :) 
Would you be open to fix that? I am away this weekend and can't do anything.

TxHawks:
Sure. I'll do it tomorrow

TxHawks:
Replaced `style` with `css`. After thinking about it again, I think you're actually right, and there is probably no need for `rule` here, so I didn't add it. We can always easily add it later if a valid use case comes up

rofrischmann:
Awesome, I'd like to merge and release today 👍

PR #565, opened 2018-05-03
No longer append unit to zero-values in unit plugin

HugoGiraudel:
## Description

There is no need to add a unit for a zero (`0`) value (unless it is a duration, which is of no concern for this plugin). This pull-request no longer add a unit for these values.

#### Patch

* `fela-plugin-unit`

## Checklist

#### Changes


rofrischmann:
Although its a major update isn't it? The output is obviously changed, so tests might fail if ppl are using ^ versions.

HugoGiraudel:
It is not breaking anything from a CSS & layout standpoint. If you’d rather bump a major, do so but I don’t think it’s necessary.

PR #570, opened 2018-06-05
Add support for custom isUnitlessProperty

jtmcgrath:
## Description

Adds an extra argument to `fela-plugin-unit` which allows users to define a custom `isUnitlessProperty` check function.

In `react-native`, `lineHeight` is a unitless value which equates to `px` in web. So that we can build native-first, we'd like to use a custom `isUnitlessProperty` function which does not include `lineHeight`; we'd then be able to have `lineHeight: 15` transformed to `lineHeight: '15px'` for example.

## Example

```js
import isUnitless from 'css-in-js-utils/lib/isUnitlessProperty'

const addUnitsTo = {
	lineHeight: true
}

const isUnitlessProperty = prop => addUnitsTo[prop]
	? false
	: isUnitless(prop)

unit(
	'px',
	{},
	isUnitlessProperty
)(style)
```

#### Minor

`fela-plugin-unit`

## Checklist

#### Changes

---

Note: `yarn check` failed, but running the four commands separately worked. Wasn't sure whether the example should include the new function; feels like it's not necessary, since it's a more advanced feature, but happy to add in if desired.

Also happy to consider different approaches - thought of a couple more, but this felt like the right place to add it!


rofrischmann:
I see that makes sense! Thanks for the PR =)

PR #572, opened 2018-06-12
fix renderer bug under safari

eenagy:
## Description
After update from next 5 to next 6, everything was working correctly except safari styling. Either two things happened:  
 

- infinite reload with webpack under development mode
- classes got assigned to other elements as well, not just components which were using it, resulting in wrong styling

## Example
I've already spent a couple of hours trying to make such a project for bug demo, but I don't consider that it worth the effort considering the simplicity of the fix.

#### Minor
`fela-bindings`

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/ProviderFactory.js
+++ b/packages/fela-bindings/src/ProviderFactory.js
@@ -4,6 +4,7 @@ import objectEach from 'fast-loops/lib/objectEach'
 
 function hasDOM(renderer) {
   return (
+    renderer &&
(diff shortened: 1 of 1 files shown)


rofrischmann:
Omg, really? That was pretty simple indeed!
Strange this is: I have similar issues with Next at the moment, but I've not even used Fela at all by now. Just prototyped with inline style...

Anyways, will merge that as soon as possible.
Thanks for your PR =)

Issue #494, opened 2018-01-05
Enzyme shallow output does not match expectations


HamidAghdaee:
**Type**: Bug 

| Packages | Version |
| -------- | ------- |
| fela     | 6.1.3   |
| enzyme     | 3.3.0   |
| fela-tools     | 5.1.2   |
| react     | 16.2.0   |
| react-fela    | 6.2.4   |
| enzyme-adapter-react-16 | 1.1.1|

-------------------
#### Description
When using enzyme shallow rendering, the output does not render Fela styles.

#### Steps to reproduce
I took the code from here: http://fela.js.org/docs/recipes/TestingFelaComponents.html
and added it to Repo: https://github.com/HamidAghdaee/test-fela

steps:
Clone Repo
yarn test

#### Expected Behavior
According to: 
http://fela.js.org/docs/recipes/TestingFelaComponents.html
the output of ```container.snapshot()``` in here: https://github.com/HamidAghdaee/test-fela/blob/master/src/Box.test.js#L8 should be:
```javascript
// Object {
//   "component": <div
//     className="a b c"
// >
//     hello
// </div>,
//   "styles": ".a{width:10px}.b{height:10px}.c{background-color:red}",
// }
```

#### Actual Behavior
instead the snapshot provided here:
https://github.com/HamidAghdaee/test-fela/blob/master/src/__snapshots__/Box.test.js.snap
shows 
```javascript
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`renders without crashing 1`] = `
<boxRules
  _felaTheme={Object {}}
>
  hello
</boxRules>
`;

exports[`renders without crashing 2`] = `""`;
```
I'm not sure if this is related but calling `render()` on the enzyme wrapper results in the following error even-though the renderer is indeed placed in the context:

```    createComponent() can't render styles without the renderer in the context. Missing react-fela's <Provider /> at the app root?
```

rofrischmann:
I guess this is due to changes in Enzyme 3 (for React 16).
It works perfectly fine if I use 2 and 15 as used in the docs. I am not really sure how to fix it for Enzyme 3 actually, but it's definitely not a bug in Fela itself.

HamidAghdaee:
@rofrischmann Dang. Thanks for looking into this. That's a bummer. We're trying to do stateful shallow testing of our components in React 16. As it is I can't think of a good solution until this is fixed. 

[...]

PR #582, opened 2018-07-05
Jest Testing Tools

rofrischmann:
## ToDo
- [x] add general jest bindings
- [x] export react/preact/inferno tools

## Description
Introducing a general tool to create snapshots that include both rendered CSS and HTML markup in a minimal, formatted output.

## Example

```javascript
import { createSnapshot } from 'jest-react-fela'

describe('Rendering Fela components', () => {
  it('should render correctly', () => {
    const snapshot = createSnapshot(
      <FelaComponent style={{ color: 'blue' }} />
    )

    expect(snapshot).toMatchSnapshot()
  })
})
```
```
.a {
  color: blue
}

<div className=a />
```

#### Major
- jest-react-fela
- jest-preact-fela
- jest-inferno-fela
- jest-fela-bindings

## Checklist

- [x] The code was formatted using Prettier (`yarn run format`)
- [x] The code has no linting errors (`yarn run lint`)
- [x] All tests are passing (`yarn run test`) 
- [x] There are no flow-type errors (`yarn run flow`)

#### Changes
If one of the following checks doesn't make sense due to the type of PR, just check them.

- [x] Tests have been added/updated
- [x] Documentation has been added/updated
- [x] My changes have proper flow-types

PR #586, opened 2018-07-10
added fela-enzyme

HamidAghdaee:
## Description
This PR provides new utilities to fix this issue:
https://github.com/rofrischmann/fela/issues/494

I have not updated this documentation: http://fela.js.org/docs/recipes/TestingFelaComponents.html

Once we land on something we agree, I'll do so.

## Example
https://github.com/HamidAghdaee/cf-weather

```javascript
import React from 'react';
import { felaShallow}  from 'fela-enzyme';
import { createComponent } from 'react-fela';

const theme = {
  fontSizes: [10, 12, 14, 16, 20, 24, 32, 48, 64, 80],
  color: {
    grass: '#9BCA3E',
    black: '#000'
  },
  fontSize: '15px'
};

const Div = createComponent(() => ({ color: 'black' }));
const { wrapper, snapshot } = felaShallow(<Div />, {}, theme);
expect(snapshot(wrapper)).toMatchSnapshot();

// Object {
//   "component": <div
//     className="a"
//   />,
//   "styles": "
// .a {
//   color: black
// }
// ",
// }

```

#### Major
fela-enzyme: 1.0.0

## Checklist

#### Changes

thanks


rofrischmann:
Yeah, I like that one! Have you seen https://github.com/rofrischmann/fela/pull/582 as well? It's a way more opinionated tool with minimal formatted output that's kind of what we're using at work nowadays. Guess the core benefit with enzyme is that you're able to simulate clicks and stuff right?

HamidAghdaee:
@rofrischmann yes. With this PR you get all the benefits of enzyme + fela rendered styles. The other solution looks elegant, it always does a deep render however, right? I believe the main benefit of my PR is the ability to do shallow rendering.

rofrischmann:
Yeah right! I guess we should merge both simultaneously and update the Testing docs together. It could be both smaller and simpler :P

HamidAghdaee:
@rofrischmann Sounds great. Whatever you see fit. Let me know how you want to proceed. Two things:

1. currently I haven't added flow types to any of my code. Is this something you'd like me to do?
2. I've made the package peer dependent on any version of react, enzyme and enzyme-to-json. This seems right to me. I don't think this package should be making explicit dependencies on these. I could however be more explicit what versions of these packages I've tested these with. But I have made explicit dependencies on fela. fela-tools and react-fela. I'm really not sure what values are correct for these dependencies. Perhaps they should also be peerdependencies.

rofrischmann:
Would be cool if you could add some Flow types to better ensure this package works fine in the future.
Let's collaborate on updating those docs, how does that sound? We could chat on Gitter or Twitter e.g.

PR #588, opened 2018-07-18
Feature CSSType

wagerfield:
## Description

React's `CSSProperties` interface extends from CSSType's `Properties` interface in exactly the same way that I have done in this PR.

Using the typings from `csstype` rather than `react` means that projects that are using Fela _but not React_ don't have to add `react` as a dependency just to get CSS typings.

### DIFF
--- a/packages/fela/index.d.ts
+++ b/packages/fela/index.d.ts
@@ -1,6 +1,6 @@
 declare module "fela" {
 
-  import { CSSProperties } from 'react';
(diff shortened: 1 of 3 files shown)


wagerfield:
Hey @rofrischmann what are your thoughts on getting this merged?

rofrischmann:
Ill merge both as soon as possible. Am not “allowed” to do computer stuff due to a inflamation in both hands

PR #599, opened 2018-08-18
Replace unallowed symbols in class name with underscore

### LINKED ISSUE #596: fela-monolithic with prettySelectors: true doesn't escape characters in displayName
**Type**: Bug

| Packages | Version |
| -------- | ------- |
| fela | ^6.1.9 |
| fela-dom | ^7.0.9 |
| fela-monolithic | ^5.0.23 |
| react-fela | ^7.3.1 |

-------------------
#### Description
`fela-monolithic` with `prettySelectors: true` doesn't escape characters in class name such as parenthesis.

#### Steps to reproduce

1. Create component
2.  Add displayName with symbols which needs to be escaped in class name (e. g. parenthesis)
2. Use createComponent to add styles

[repository with minimal reproducable example](https://github.com/Nitive/fela-display-name-bug)

#### Expected Behavior
displayName is escaped

#### Actual Behavior
displayName is not escaped

Nitive:
Fixes #596

## Description
Replace unallowed symbols with underscore for class names based on function `.name` or component `.displayName`.

I wanted to escape class names at first, but this approach has two downsides:
1. We need to add `css.escape` dependency to polyfill CSS.escape. It would be used for development only, but still
2. We need to escape it in style tag, but keep unescaped name in dom attribute. It adds extra troubles

#### Patch

- fela-bindings
- react-fela
- preact-fela
- inferno-fela

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/__tests__/__snapshots__/createComponentFactory-test.js.snap
+++ b/packages/fela-bindings/src/__tests__/__snapshots__/createComponentFactory-test.js.snap
@@ -889,6 +889,48 @@ Array [
 ]
 `;
 
+exports[`Creating Components with a Proxy for props from Fela rules should replace rare unallowed symbols in className with underscore 1`] = `
+Array [
+  "<style>
+    .rule_1_______________abrv9k {
+        font-size: 16
+    }
+</style>",
+  <rule>
+    <rule
+        _felaTheme={Object {}}
+    >
+        <1!@#$%^&*{}/=\\
+            className="rule_1_______________abrv9k"
+        >
+            <div />
+        </1!@#$%^&*{}/=\\>
+    </rule>
+</rule>,
+]
+`;
+
+exports[`Creating Components with a Proxy for props from Fela rules should replace unallowed symbols in className with underscore 1`] = `
+Array [
+  "<style>
+    .rule_connect_Component___abrv9k {
+        font-size: 16
+    }
+</style>",
+  <rule>
+    <rule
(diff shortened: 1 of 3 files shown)


packages/fela-bindings/src/__tests__/__snapshots__/createComponentFactory-test.js.snap
```diff
@@ -889,6 +889,48 @@ Array [
...
+    .rule_connect_Component___abrv9k {
+        font-size: 16
+    }
+</style>",
+  <rule>
+    <rule
+        _felaTheme={Object {}}
+    >
+        <connect(Component)
+            className="rule_connect_Component___abrv9k"
```
Nitive: Readable class name in common cases

packages/fela-bindings/src/__tests__/__snapshots__/createComponentFactory-test.js.snap
```diff
@@ -889,6 +889,48 @@ Array [
 ]
 `;
 
+exports[`Creating Components with a Proxy for props from Fela rules should replace rare unallowed symbols in className with underscore 1`] = `
+Array [
+  "<style>
+    .rule_1_______________abrv9k {
```
Nitive: Unique, but unreadable class name, because most of the displayName contains unallowed symbols (`displayName = '1!@#$%^&*{}/=\\'`)

rofrischmann:
Thanks for the PR :)

PR #601, opened 2018-08-21
Handle exceptions extracting used props

### LINKED ISSUE #595: createComponentWithProxy: extractUsedProps may throw an unhandled exception
**Type**: Bug

| Packages | Version |
| -------- | ------- |
| fela-bindings     | 1.3.4   |

-------------------
#### Description

Error rendering a component created with `createComponentWithProxy`, when the passed rule throws an exception if one of the properties is undefined. I think the problem is that when extracting used properties, the rule is invoked passing an empty props object, so this [line](https://github.com/rofrischmann/fela/blob/master/packages/fela-bindings/src/extractUsedProps.js#L27) can throw an unhandled exception, if the rule doesn't handle undefined values. I tried wrapping that line with a try/catch and it seems to work.

#### Steps to reproduce

https://codesandbox.io/s/jz79wq325y

astrada:
Fix #595

## Description
Add a try/catch in `extractUsedProps` when the rule is evaluated. If the rule throws an unhandled exception, returns an empty array.

#### Patch

* fela-bindings

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/__tests__/extractUsedProps-test.js
+++ b/packages/fela-bindings/src/__tests__/extractUsedProps-test.js
@@ -1,0 +1,19 @@
+import extractUsedProps from '../extractUsedProps'
+
+describe('Extracting used props', () => {
+  it('should return used props', () => {
+    expect(
+      extractUsedProps(props => {
+        props.foo, props.bar
+      })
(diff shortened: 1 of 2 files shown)


rofrischmann:
Thanks for the PR!

PR #607, opened 2018-09-15
Update ProviderProps in index.d.ts

dlee:
## Description
`ProviderProps` did not match what `Provider` was actually expecting. This PR updates the definition of `ProviderProps` to remove `mountNode` and add `rehydrate`.


rofrischmann:
Thanks for the PR :)

Issue #538, opened 2018-03-18
createComponent wrapping createComponentWithProxy passes through unwanted props


dlee:
**Type**: Bug

| Packages | Version |
| -------- | ------- |
| fela     | 6.0.x   |

-------------------
#### Description
Wrapping a component created by `createComponentWithProxy` with `createComponent` bleeds all properties into the DOM element, regardless of the `passThroughProps` declared in `createComponent`

#### Steps to reproduce
```javascript
const BorderlessButton = createComponentWithProxy(() => ({
  border: 'none'
}), 'button');

const ColoredButton = createComponent(({color}) => ({
  color
}), BorderlessButton);

render(
  <ColoredButton color="red" />
);
```

#### Expected Behavior
```html
<button class="a b"></button>
```

#### Actual Behavior
```html
<button color="red" class="a b"></button>
```

rofrischmann:
Alright, these HoCs are driving me mad 😅
I'm experimenting with some major changes that will reduce the complexity and issues with those by simply replacing all HoCs entirely (providing a codemod of course).

kserjey:
When you plan to release that?

rofrischmann:
Alright, so we decided to not fix this one as it creates issues with other use cases.
We recommend not wrapping both HoCs together at all or using FelaComponent.

We will release improved APIs sometime in November.

PR #611, opened 2018-09-21
Stop leaking composed passThroughProps with createComponentWithProxy

rofrischmann:
## Description
This will fix a common issue when composing components created with `createComponentWithProxy` (https://github.com/rofrischmann/fela/issues/538)

## Example

```javascript
const BorderlessButton = createComponentWithProxy(() => ({
  border: 'none'
}), 'button');

const ColoredButton = createComponent(({color}) => ({
  color
}), BorderlessButton);

render(
  <ColoredButton color="red" />
);
```

will now yield 
```HTML
<button class="a b"></button>
```

rather than
```HTML
<button color="red" class="a b"></button>
```

## Problem

The problem is, with this the following won't work either:
```js
const ColoredButton = createComponent(({color}) => ({
  color
}), BorderlessButton, ['color']);
```
because `createComponentWithProxy` will ignore composed passThroughProps.

#### Patch
* fela-bindings

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/__tests__/__snapshots__/createComponentFactory-test.js.snap
+++ b/packages/fela-bindings/src/__tests__/__snapshots__/createComponentFactory-test.js.snap
@@ -42,8 +42,16 @@ Array [
         onClick={[Function]}
     >
         <FelaComponent
+            _felaProps={
+                Object {
+                    "as": undefined,
+                    "color": "red",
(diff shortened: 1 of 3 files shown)


rofrischmann:
I've thought about it more and more and I think this is not worth the overhead. One should not mix multiple different createComponent calls.

PR #616, opened 2018-10-02
Separate Typescript definitions

rtheone:
## Description
The Typescript compiler throws a dependency error when not using React.

```
C:\Generic Fela-Inferno Project
> tsc
node_modules/fela/index.d.ts:298:26 - error TS2307: Cannot find module 'react'.

298   import * as React from "react";
```

To resolve this, the React-relevant definitions have been moved to the `react-fela` package.
In addition, definitions for Inferno (based on the former) have been added to `inferno-fela`.

With some cleanup and additions, Fela's definitions should be considered potentially viable
for submission to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) to become available on `npm`  via the `@types/*` scope.

## Checklist
No changes were made to the core library itself. With the changes, Typescript compilation 
of personal Inferno and React Fela apps seem to complete without issue.


rofrischmann:
Thanks for the PR @rtheone! I'd love to see the typings being added to DefninitelyTyped. Would be up for sth. like that?

PR #617, opened 2018-10-05
Export felaMount from fela-enzyme

UberMouse:
This seems unintentional? I have 42 failing tests which I doubt this change could have affected so I'm not sure what's going on there.

#### Patch
* fela-enzyme

## Checklist

#### Changes

---
This change is [Reviewable](https://reviewable.io/reviews/rofrischmann/fela/617)


rofrischmann:
Thanks for adding that @UberMouse! I have no deep knowledge of the fela-enzyme package, but I guess that one should be exposed yeah. We might have @HamidAghdaee helping here.

PR #619, opened 2018-10-07
add badges for bundlephobia in all packages

wcastand:
## Description
Add bundlephobia badges to all packages. (some package seems to not be working because of dependencies error. example : [https://bundlephobia.com/result?p=fela-layout-debugger@latest](fela-layout-debugger)

## Versioning
i just updated the readme for each package, so a patch or minor should be good no ?

#### Quality Assurance

I didn't touch the code at all, only the readme so not sure it's relevant.
> You can also run `yarn run check` to run all 4 commands at once.

- [ x ] The code was formatted using Prettier (`yarn run format`)

#### Changes


rofrischmann:
Thanks for adding those! Actually we don't have to release anything as it's just "docs".

PR #622, opened 2018-10-09
change to minzip version bundle and add fela core package to main readme

wcastand:
Like the title said, i changed to minzip instead of min and added core package bundle size to main readme.

related to #619


rofrischmann:
awesome!

Issue #589, opened 2018-07-18
react-fela: componentWillMount is unsafe React.StrictMode


mfdj:
**Type**: ~Bug~ Improvement 

| Packages | Version |
| -------- | ------- |
| react-fela     | 7.1.0   |
| fela     | 6.1.5   |

-------------------
#### Description

Promoting this comment to an issue https://github.com/rofrischmann/fela/issues/532#issuecomment-380403527:

> As per 16.3, componentWillMount is entering deprecation phase in favor of componentDidMount.
>
> Right now, it means it is not realistically possible to use React.StrictMode in a codebase using Fela because it produces a swarm of warnings.

#### Steps to reproduce

1. Enable [React.StrictMode](https://reactjs.org/docs/strict-mode.html)
2. Implement a component using `createComponent` from react-fela
3. Build the app

#### Expected Behavior
No errors output.

#### Actual Behavior
Errors will be output:
```
"Warning: Unsafe lifecycle methods were found within a strict-mode tree:

componentWillMount: Please update the following components to use componentDidMount instead: FelaComponent
```

rofrischmann:
It's not actually a bug, but a known issue with never React versions. Anyways, it should work as expected for now. We're working on a rework of all the bindings to support newer React features soon :)

PR #623, opened 2018-10-09
migration on new react lifecycle

sgrishchenko:
## Description
This pull request should help migrate to the use of a new react life cycle.
(for https://github.com/rofrischmann/fela/issues/589)

#### Patch
fela-bindings
fela-plugin-native-media-query

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/ThemeProviderFactory.js
+++ b/packages/fela-bindings/src/ThemeProviderFactory.js
@@ -20,9 +20,9 @@ export default function ThemeProviderFactory(
       this.theme = createTheme(props.theme, previousTheme)
     }
 
-    componentWillReceiveProps(nextProps: Object): void {
-      if (!shallowEqual(this.props.theme, nextProps.theme)) {
-        this.theme.update(nextProps.theme)
+    componentDidUpdate(prevProps: Object): void {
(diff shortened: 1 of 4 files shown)


packages/fela-bindings/src/ThemeProviderFactory.js
```diff
@@ -20,9 +20,9 @@ export default function ThemeProviderFactory(
       this.theme = createTheme(props.theme, previousTheme)
     }
 
-    componentWillReceiveProps(nextProps: Object): void {
-      if (!shallowEqual(this.props.theme, nextProps.theme)) {
-        this.theme.update(nextProps.theme)
+    componentDidUpdate(prevProps: Object): void {
```
rofrischmann: Doesn't this eventually cause an additional update? Before we updated the theme before the actual update happens. Now we update it after the props change => another update.

sgrishchenko: I think this code is equivalent. We update the theme to update all the components that are subscribed to it. Theme update = re-render subscribers. Therefore, there is no difference in the re-rendering of subscribers before the provider has been updated or after.

PR #624, opened 2018-10-09
fix type error

wcastand:
I just realized i made a typo when i use find/replace :/ sorry about that


rofrischmann:
:D wasn't even checking the code whoops.

PR #645, opened 2018-12-14
Fix theme providers not re-rendering their children

troch:
## Description

`shouldComponentUpdate` on theme providers prevent them to re-render their children if the provided theme hasn't changed. I understand it might have been added to prevent re-rendering all styled fela components if the theme hasn't changed, but React context uses reference identity to determine when to re-render. After checking, [create-inferno-context](https://github.com/kurdin/create-inferno-context/blob/master/src/implementation.js#L67-L69) and [preact-context](https://github.com/valotas/preact-context/blob/master/src/context.ts#L87) also have the same behaviour.

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/package.json
+++ b/packages/fela-bindings/package.json
@@ -28,8 +28,7 @@
     "fast-loops": "^1.0.0",
     "fela-dom": "^10.0.1",
     "fela-tools": "^10.0.1",
-    "react-addons-shallow-compare": "^15.6.2",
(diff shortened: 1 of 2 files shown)


rofrischmann:
Would you consider it a patch release? Seems to me that nothing breaks as the new Context API supports that check out of the box.

troch:
Yep, patch

troch:
Thanks a lot for merging and releasing 🙇 !

PR #658, opened 2019-01-20
fix(fela-plugin-extend): don't throw on undefined items in extend array

### LINKED ISSUE #657: fela-plugin-extend crushes on undefined
**Type**: Bug

| Packages | Version |
| -------- | ------- |
| fela     | 6.1.7   |
| react-fela     | 7.2.0   |
| fela-plugin-extend     | 6.0.11   |

-------------------
#### Description
Passing `undefined` to the `extend` array throws a reference error about `hasOwnProperty of undefined`.

#### Steps to reproduce
```js
const style = theme => ({
  extend: [ undefined, ],
});
```

#### Expected Behavior
`undefined` should just be discarded

TxHawks:
## Description
Fixes #657 where `fela-plugin-extend` would throw when the `extend` key in a style object was an array with a `null` or `undefined` item inside it.

Adds a test

## Packages

fela-plugin-extend

## Versioning

- [ ] Major
- [ ] Minor
- [x] Patch

## Checklist

#### Changes

### DIFF
--- a/packages/fela-plugin-extend/src/index.js
+++ b/packages/fela-plugin-extend/src/index.js
@@ -22,7 +22,10 @@ function removeUndefined(style: Object): Object {
 
 function extendStyle(style: Object, extension: Object): void {
   // extend conditional style objects
-  if (extension.hasOwnProperty('condition')) {
+  if (
+    extension &&
+    Object.prototype.hasOwnProperty.call(extension, 'condition')
+  ) {
     if (extension.condition) {
       // eslint-disable-next-line no-use-before-define
       assignStyle(style, extend(extension.style))
--- a/packages/fela-plugin-extend/src/__tests__/extend-test.js
+++ b/packages/fela-plugin-extend/src/__tests__/extend-test.js
@@ -156,6 +156,18 @@ describe('Extend plugin', () => {
     })
   })
 
+  it('should filter out null items in extend array', () => {
+    const base = {
+      color: 'blue',
+      extend: [{ backgroundColor: '#ccc' }, null],
+    }
+
+    expect(extend()(base)).toEqual({
+      color: 'blue',
+      backgroundColor: '#ccc',
+    })
+  })
+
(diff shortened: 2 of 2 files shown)


packages/fela-plugin-extend/src/index.js
```diff
@@ -22,7 +22,10 @@ function removeUndefined(style: Object): Object {
 
 function extendStyle(style: Object, extension: Object): void {
   // extend conditional style objects
-  if (extension.hasOwnProperty('condition')) {
+  if (
+    extension &&
+    Object.prototype.hasOwnProperty.call(extension, 'condition')
```
rofrischmann: What's the benefit of calling it this way instead of just `extension.hasOwnProperty('condition')`?

TxHawks: It's not a big deal, but if someone uses `Object.create(null)` to create a lightweight map without the entire object prototype, `.hasOwnProperty` would not exist and the code would throw an `hasOwnProperty is not a function` TypeError.

So, if somene did:
```js
<FelaComponent 
  style={Object.assign(Object.create(null), { extend: [ /* ... */ ], })}
>
```

that would error out.

I wouldn't necessarily bother with this kind of thing in app code, but I feel library code should be as robust as possible and I don't really see any down side to it.

TxHawks: Let me know if you'd like me to change it back or release it like this

rofrischmann: Oh sure not, that makes sense - didn't know it =)

TxHawks:
Can't release because flow errors out on unrelated issues

rofrischmann:
Hmmmm, experienced the same issue with Flow in another repo and I couldn't even ignore them. Feel free to disable flow for the release

TxHawks:
updated the changelog, but I can't merge it without you approving it

PR #667, opened 2019-01-31
feat: add `rendererId` field to config

### LINKED ISSUE #664: `devMode` is broken with multiple renderers
**Type**: Bug

| Packages | Version |
| -------- | ------- |
| `fela`     | 10.1.3   |
| `react-fela`     | 10.1.3   |

-------------------
#### Description
`devMode` option breaks multiple renderers.

#### Steps to reproduce
One example is worth a thousand words 😸 
https://codesandbox.io/s/711kjkr1o6

#### Expected Behavior
Everything works.

#### Current Behavior
Styles are applied only for second header.

layershifter:
Fixes #664.

## Description
Adds `rendererId` field to config that we will be used in animation names and will be added `<style>` as `data-fela-id` if present.

## Example
image

```js
createRenderer({ rendererId: 'awesome' })
```

## Packages
- `fela`
- `fela-dom`

## Versioning

- [ ] Major
- [x] Minor
- [ ] Patch

## Checklist

#### Changes


flowtypes/DOMRenderer.js
```diff
@@ -4,12 +4,12 @@ export type DOMRenderer = {
   keyframePrefixes: Array<string>,
   plugins: Array<Function>,
   mediaQueryOrder: Array<string>,
+  rendererId: string,
   selectorPrefix: string,
   filterClassName: Function,
   listeners: Array<Function>,
   uniqueRuleIdentifier: number,
   uniqueKeyframeIdentifier: number,
-  isConnectedToDOM: boolean,
```
layershifter: Cruft, is used nowhere 👾

docs/advanced/RendererConfiguration.md
```diff
@@ -12,6 +12,7 @@ We might introduce more configuration options with future releases, so be sure t
 | keyframePrefixes | *(Array?)* |`['-webkit-',`<br>`'-moz-']` | A list of which additional `@keyframes` prefixes are rendered |
 | enhancers  | *(Array?)* |  |  A list of [enhancers](../advanced/Enhancers.md) to enhance the renderer
 | mediaQueryOrder | *(Array?)* |  | An explicit order in which `@media` queries are rendered |
+| rendererId | *(string?)* |  | An optional unique identifier that will prefix your animation names and will be used in `<style>` nodes. |
```
TxHawks: I feel this isn't clear enough.

Why would someone use this?  What is the use case? Why might someone want to perfix their animation names and what's the connection between animation names and the attribute on the style nodes?

Without knowing in advance, I couldn't have figured it out from this description.

Also, I think it's be clearer if it were `will be added as the value of a data-fela-id attribute on <style> nodes` rather than `used in <style> nodes`

packages/fela/README.md
```diff
@@ -1,237 +1,237 @@
-<h1><img alt="Fela" src="docs/res/logo.png" width="203"/></h1>
-
```
TxHawks: For some reason the whole file is marked as changed. Was it actually edited? I failed to find the changes

packages/fela/src/generateAnimationName.js
```diff
@@ -1,4 +1,7 @@
 /* @flow */
-export default function generateAnimationName(id: number): string {
-  return `k${id}`
+export default function generateAnimationName(
+  rendererId: string,
+  id: number
+): string {
+  return `${rendererId}k${id}`
```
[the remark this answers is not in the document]
layershifter: It was added, https://github.com/rofrischmann/fela/pull/667/files#diff-1b838cf86eb6f3199253f12209e04222

docs/advanced/RendererConfiguration.md
layershifter: This is much cleaner 👍 Updated

packages/fela/README.md
layershifter: This file had CRLF line endings I accidentally touched it and now they are LF
Will try to revert these changes

layershifter: Reverted 👍

packages/fela/src/generateAnimationName.js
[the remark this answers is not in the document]
TxHawks: Sorry, I missed that

docs/advanced/RendererConfiguration.md
TxHawks: Thanks. It's clear now _what_ it does, but still not really clear what one would use that for.

I think it'd be good to describe the use case

layershifter: > An example use case: the recommended way is to have only one renderer, but there are some specific cases when you will need to have multiple renderers. This option allows you avoid conflicts in generated CSS rules via prefixing animation names and separated `<style>` tags.

Going to add this? Do you have any suggestions?

TxHawks: Yeah, that's great. Thanks

layershifter: Pushed 👍

layershifter:
@rofrischmann can you please take a look? 🐱

packages/fela/README.md
rofrischmann: This file is auto-generated/copied on release, so it doesn't matter anyways :P

docs/advanced/RendererConfiguration.md
```diff
@@ -32,7 +33,7 @@ import { renderToString } from 'fela-dom'
 const config = {
   plugins: [ unit('em'), prefixer(), fallbackValue() ],
   keyframePrefixes: ['-webkit-'],
-  enhancers: [ beautifer() ],
+  enhancers: [ beautifier() ],
```
rofrischmann: extra points for that one :D

packages/fela/src/generateAnimationName.js
```diff
@@ -1,4 +1,7 @@
 /* @flow */
-export default function generateAnimationName(id: number): string {
-  return `k${id}`
+export default function generateAnimationName(
+  rendererId: string,
```
rofrischmann: Can we switch position here? I'm not sure if anyone uses that specific function separately, but it'd be nice to break it if the rendererId is missing. Also set a default param value of `''`

packages/fela-dom/src/dom/connection/createNode.js
```diff
@@ -5,6 +5,7 @@ import type { NodeAttributes } from '../../../../../flowtypes/DOMNode'
 
 export default function createNode(
   nodes: Object,
+  id: string,
```
rofrischmann: Same ordering thing here. Would like to keep it consistent and only add params at the end.

rofrischmann:
I added some small notes here.
Also, there's one last important thing we need to adjust:
https://github.com/rofrischmann/fela/blob/82be80528fa9e3b85e7fdb84645b04d670eb4e38/packages/fela-dom/src/dom/rehydrate.js#L17

The rehydration function should now compare data-fela-id and renderer.rendererId (passed renderer) to only rehydrate styles from that specific renderer. It wouldn't crash otherwise, but result in mixed styles and a lot of unnecessary cache entries.

packages/fela/src/generateAnimationName.js
layershifter: Updated 👍

packages/fela-dom/src/dom/connection/createNode.js
layershifter: Updated 👍

layershifter:
> Last but not least, we'd have to update `renderToSheetList` as well in order to have both SSR methods fully working.

Can you please clarify what updates you're expecting? :)

```jsx
const firstRenderer = createRenderer({ rendererId: 'FIRST' })
const secondRenderer = createRenderer({ rendererId: 'FIRST' })

firstRenderer .renderRule(rule)
secondRenderer.renderRule(rule)

const firstSheetList = renderToSheetList(renderer)
// OUTPUT:
// { type: 'RULE', css: '.a{font-size:12px}.b{color:blue}', rehydration: 4 },
const secondSheetList = renderToSheetList(renderer)

firstNode.append(firstSheetList)
secondNode.append(secondSheetList)
```

I _know_ which renderer I'm using and I know where put styles. I'm not sure but it seems that we don't need to update anything there.

rofrischmann:
Hmm, you're right. But I feel like it might be useful to pass the id for each sheet as well, so we can just iterate the attributes and add them to the nodes.

layershifter:
> But I feel like it might be useful to pass the id for each sheet as well, so we can just iterate the attributes and add them to the nodes.

Yep, but `rendererId` is optional field. So as consumer to get it usefull I should always define it explicitly:
```jsx
const firstRenderer = createRenderer({ rendererId: 'FIRST' })
const secondRenderer = createRenderer({ rendererId: 'SECOND' })
```

It will also cause complications in code of `renderToSheetList`:

### Evil option 1

```diff
-list.push({
-  css: cacheCluster[key],
-  rehydration: rehydrationIndex,
-  type,
-})
+const sheet = {
+  css: cacheCluster[key],
+  rehydration: rehydrationIndex,
+  type,
+}
+
+if (renderer.rendererId) {
+  sheet.rendererId = renderer.rendererId
+}
+
+ list.push(sheet)
```

We do 4 `.push()` times, will look 💩 

### Evil option 2

```diff
+const assignRendererId = (sheet, rendererId) => {
+  if (rendererId) {
+    sheet.rendererId = rendererId
+  }
+}
//
list.push(
+ assignRendererId({
    css: cacheCluster[key],
    rehydration: rehydrationIndex,
    type,
+ }, renderer.rendererId),
)
```

Not so 💩 , but will add a ton of function calls.

rofrischmann:
Well hmm. Then we should at least update the Server-side Rendering docs and/or renderToSheetList docs to reflect that it is required to add that attribute.

layershifter:
@rofrischmann I've added a note to `renderToMarkup.md`, please check. Is it enough?

rofrischmann:
That one looks fine yeah. Now `renderToSheetList` is the only one missing. Basically making sure that when using it, people also pass the `data-fela-id` attribute to their style nodes.

layershifter:
Added a note to `renderToSheetList.md` about `data-fela-id` 👍

rofrischmann:
LGTM =) will merge and publish tomorrow

rofrischmann:
@layershifter Again, thanks for all the effort here and all the changes etc. :P

PR #671, opened 2019-02-12
fix(fela): get rid of const enum in typings

layershifter:
## Description

Currently `fela` typings have a single use of `const enum` and it means that you cannot to build a project with `--isolatedModules`. See for more details: https://github.com/stardust-ui/react/issues/613

Actually, this change is very important because `const enum`s are not supported by `@babel/preset-typescript`, too.

I added typings for `fela-utils` as this package contains constants that can be used by consumers.

> You should use a union type of literals (string or number) instead.

https://github.com/Microsoft/TypeScript/issues/20703#issuecomment-362005736

## Packages
- `fela`

## Versioning
Actually, it's a breaking change that requires a major bump because you will not able to use `TSubscribeMessageType` as value.

## Before

```ts
import { TSubscribeMessageType } from 'fela'

const type = TSubscribeMessageType.rule
```

## After
```ts
import { RULE_TYPE } from 'fela-utils'

const type = RULE_TYPE 
```

## Checklist

### DIFF
--- a/packages/fela/index.d.ts
+++ b/packages/fela/index.d.ts
@@ -1,5 +1,6 @@
 declare module "fela" {
   import * as CSS from 'csstype';
+  import { TRuleType, TKeyframeType, TFontType, TStaticType, TClearType } from 'fela-utils';
 
   export type TRuleProps = {};
   export type TRule<T = TRuleProps> = (props: T) => IStyle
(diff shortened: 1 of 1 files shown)


rofrischmann:
Hmm so we should have a major version update "just" for those types? Feels somewhat wrong to me, but makes sense...

layershifter:
> Hmm so we should have a major version update "just" for those types?

For me: it's fine to ship it as patch 😸

However, if someone really used `TSubscribeMessageType` as value, it will be a breaking change 🤔 Have mixed feelings about it.

rofrischmann:
Yeah me too. But as someone who does not use TypeScript, this feels super wrong as the actual code is not breaking or anything.
Let's try a patch and see what happens - maybe we learn it should be done with majors next time 🙆‍♂️

PR #674, opened 2019-02-19
perf(useFela): optimize passing of theme & props

TxHawks:
## Description

This pull requests makes the following suggestions

#### Optimizations

In many simple use cases, the `useFela` hook will be used without any props passed to it, and will only make the `theme` available to the rule callbacks in the returned `css` function. In light of that,
this commit makes two optimizations:

1. Only merge props into the object that is passed to `css`'s callbacks when props are actually passed.
2. Merge props into the `propsWithTheme` object using mutative `Object.assign`, as it is by far the most performant way of doing so. See: https://jsperf.com/object-assign-vs-spreading/3

These may seem like micro optimizations, but since useFela is a hot
path that could potentially run thousands of times, I think this is
of value in this case.

#### DX

The `useFela` hook now throws an error with a helpful message when it is used in a component that isn't a descendant of a `RendererProvider`

## Packages
react-fela

## Versioning
Patch

## Checklist

#### Changes

### DIFF
--- a/packages/react-fela/src/useFela.js
+++ b/packages/react-fela/src/useFela.js
@@ -15,6 +15,12 @@ export default function useFela(props: Object = {}): HookInterface {
   const renderer = useContext(RendererContext)
   const theme = useContext(ThemeContext) || {}
 
+  if (!renderer) {
+    throw new Error(
+      'The "useFela" hook can only be used  inside a "RendererProvider"'
+    )
+  }
+
   const propsWithTheme = { theme }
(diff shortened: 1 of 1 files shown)


TxHawks:
Not sure why tests are failing in the CI env, they all pass locally

TxHawks:
Sorry, my bad. Fixed now

TxHawks:
I hate mutating like that too, but it's an object we create and completely control, so the perf gains in a hot-path seem worth it.

PR #675, opened 2019-02-19
feat(FelaComponent): Allow style prop with empty value

TxHawks:
## Description
Since v10,  FelaComponent throws if the `style` prop is `null` or `undefined`, which I think is wrong. 

It's quite an edge case, but sometimes one may want an element to only be styled under certain conditions, and have that be determined dynamically. 

The easiest and most straightforward way to do that is to pass `null` to `FelaComponent`'s style prop from the parent component.

By removing the check and `throw` statement, it works as expected, so I don't really see any benefit in having it there. It doesn't really add any DX value, and is overly restrictive

## Example
```jsx
function MaybeStyled({ style, children, }) {
  return (<FelaComponent style={style}>{children}</FelaComponent>);
}


<MaybeStyle style={null} /> // renders <div></div>
<MaybeStyle style={{ color: 'red', }} /> // renders <div class="a"></div>
```

## Packages
fela-bindings

- 

## Versioning
Patch

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/FelaComponentFactory.js
+++ b/packages/fela-bindings/src/FelaComponentFactory.js
@@ -34,12 +34,6 @@ export default function FelaComponentFactory(
       'The `render` prop in FelaComponent is deprecated. It will be removed in react-fela@11.0.0.\nPlease always use `children` instead. See http://fela.js.org/docs/api/bindings/fela-component'
     )
 
-    if (!style) {
-      throw new Error(
-        'A valid `style` prop must be passed to FelaComponent in order to render.\nSee http://fela.js.org/docs/api/bindings/fela-component'
-      )
-    }
-
     const renderFn = renderer =>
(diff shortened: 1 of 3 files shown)


rofrischmann:
The point was that you don't need a wrapping FelaComponent without a style as you can simply render a div as well. But fair enough, it seems like a legit use case to pass undefined / null.

TxHawks:
> but maybe we should log a warning in devMode?

Happy to do this. Did you release already?

rofrischmann:
Not yet, but was about to do that soonish. But I'll have another release at the end of the week anyways :P

TxHawks:
Actually, how would you know if we are in dev mode inside FelaComponent?

rofrischmann:
`renderer.devMode`

TxHawks:
Work just got really crazy, I don't think I'll manage to get to it today. Let's jus make sure to do this later

PR #679, opened 2019-02-22
Fixes exception when rendererId is not defined

gtothesquare:
This applies to the edge case in case you are using old fela with
new fela dom version.

## Description
Add a description here.

## Example

```javascript
// your code 
```

## Packages

- fela-dom

## Versioning
None / Patch / Minor / Major

## Checklist

#### Changes

### DIFF
--- a/packages/fela-dom/src/dom/rehydrate.js
+++ b/packages/fela-dom/src/dom/rehydrate.js
@@ -18,7 +18,7 @@ export default function rehydrate(renderer: DOMRenderer): void {
   render(renderer)
 
   const idQuery =
-    typeof renderer.rendererId !== 'undefined' && renderer.rendererId.length > 0
+    renderer.rendererId && renderer.rendererId.length > 0
       ? `[data-fela-id="${renderer.rendererId}"]`
       : ''
 
--- a/packages/fela-dom/src/dom/rehydrate.js
+++ b/packages/fela-dom/src/dom/rehydrate.js
@@ -18,7 +18,7 @@ export default function rehydrate(renderer: DOMRenderer): void {
   render(renderer)
 
   const idQuery =
-    renderer.rendererId.length > 0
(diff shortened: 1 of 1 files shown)


packages/fela-dom/src/dom/rehydrate.js
```diff
@@ -18,7 +18,7 @@ export default function rehydrate(renderer: DOMRenderer): void {
   render(renderer)
 
   const idQuery =
-    renderer.rendererId.length > 0
+    typeof renderer.rendererId !== 'undefined' && renderer.rendererId.length > 0
```
TxHawks: Thanks for the PR

Why do we need `typeof` here? it's not a variable that can cause a reference error. Isn't it enough to just do:
```suggestion
    renderer.rendererId && renderer.rendererId.length > 0
```

gtothesquare: True, old habit since when I saw Id I thought also it could be an int. I've made the change.

packages/fela-dom/src/dom/rehydrate.js
```diff
@@ -18,7 +18,7 @@ export default function rehydrate(renderer: DOMRenderer): void {
   render(renderer)
 
   const idQuery =
-    renderer.rendererId.length > 0
+    renderer.rendererId && renderer.rendererId.length > 0
       ? `[data-fela-id="${renderer.rendererId}"]`
       : ''
```
layershifter: See #681, we should always add `data-fela-id`

rofrischmann: We still need it for the edge case where ppl use different versions. Shouldn't happen but might be^^

rofrischmann:
So @gtothesquare Ill close this PR (see the comments) as the problem is already fixed with another PR solving it in general.

PR #681, opened 2019-02-25
fix(renderer): apply `data-fela-id` always

layershifter:
## Description
We should always add `data-renderer-id` 🤔 

## Example
Easiest repro with broken styles in `devMode` (#664): https://codesandbox.io/s/2zz49q92xj
It will be produced when a renderer with defined `rendererId` will be first.

```js
const ltrSelector = '[data-fela-type="RULE"][data-fela-id="rtl"]:not([data-fela-support="true"]):not([media])'
const rtlSelector = '[data-fela-type="RULE"]:not([data-fela-support="true"]):not([media])'

console.log(document.querySelector(ltrSelector) === document.querySelector(rtlSelector)) // true
```

After this fix we will add `data-fela-id` to all nodes:

```js
const firstSelector = '[data-fela-type="RULE"][data-fela-id=""]:not([data-fela-support="true"]):not([media])'
const secondSelector = '[data-fela-type="RULE"][data-fela-id="rtl"]:not([data-fela-support="true"]):not([media])'

console.log(document.querySelector(ltrSelector) === document.querySelector(rtlSelector)) // false
```

## Packages
-  fela-dom

## Versioning
Patch

## Checklist

#### Changes

### DIFF
--- a/packages/fela-dom/src/dom/__tests__/__snapshots__/rehydrate-test.js.snap
+++ b/packages/fela-dom/src/dom/__tests__/__snapshots__/rehydrate-test.js.snap
@@ -93,7 +93,7 @@ exports[`Rehydrating from DOM nodes should rehydrate with correct "rendererId" 1
 "<html>
 
 <head>
-  <style type=\\"text/css\\" data-fela-rehydration=\\"2\\" data-fela-type=\\"RULE\\">
+  <style type=\\"text/css\\" data-fela-rehydration=\\"2\\" data-fela-type=\\"RULE\\" data-fela-id=\\"\\">
     .a {
       color: yellow
     }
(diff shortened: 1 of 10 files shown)


rofrischmann:
Will merge that in as well for now, but do you think we could at some point deprecate the functionality again? Having everything theme-centric feels way more right to me at this point.

layershifter:
I'm going to test the latest features with `theme.direction`, but I think that we will still require two separate renderers because we have a case with RTL and LTR layouts on the same page.

TxHawks:
That's not a problem, with #686 themes are merges again, so you can have the direction changed on a per element basis simply by wrapping it with a theme provider with nothing but `{ direction: altDirection }`

layershifter:
Sounds interesting, will test it 👍 

rtl-animation

If this behavior will be preserved this will solve all our issues.

PR #686, opened 2019-03-06
Adding lost ability to auto-merge & overwrite themes

rofrischmann:
## Description
This PR simply reimplements the old behaviour that was lost when moving on to the new context API. I would consider that a bug and thus just a patch release.

## Packages

- fela-bindings
- react-fela
- preact-fela
- inferno-fela

## Versioning
Patch 

## Checklist

#### Changes

### DIFF
--- a/packages/fela-bindings/src/ThemeProviderFactory.js
+++ b/packages/fela-bindings/src/ThemeProviderFactory.js
@@ -5,26 +5,20 @@ export default function ThemeProviderFactory(
   BaseComponent: any,
   ThemeContext: any,
   createElement: Function,
-  renderChildren: Function,
-  statics?: Object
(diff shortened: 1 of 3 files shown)


packages/fela-bindings/src/ThemeProviderFactory.js
```diff
@@ -5,26 +5,20 @@ export default function ThemeProviderFactory(
   BaseComponent: any,
```
TxHawks: `baseComponent` and `the objectEach` import don't seem to be needed an more

rofrischmann: Fair enough, I will remove it before pushing. This is kind of a hot fix as I need this functionality now :D

Issue #688, opened 2019-03-10
New option to isolate.


houd1ni:
The isolate plugin: https://github.com/rofrischmann/fela/tree/master/packages/fela-plugin-isolation

I suggest to add an option or another plugin that adds some tunable prefix to fela's class names instead of adding the whole rule for each of them. E.g. 'f-a f-b f-ae' instead of making 3 copies of `all:initial`. That seems to be quite less code. I also can make a PR after negotiation.

rofrischmann:
Could you gimme a quick example of how the rendered CSS and class names would look like and how you'd want to implement that?

houd1ni:
Sure. First, it should be an option, 'cause it does not drop all previous styles, but only avoids class names collisions, that is often very close to the plugin target.

For example, one has picked `fela-` as a prefix.

### Input
```javascript
{
  color: 'green',
  fontWeight: 'bold'
}
```

### Classnames, renderRule() output:
```
fela-a fela-b
```

### Rendered CSS (beauty):
```css
.fela-a {
  color: 'green';
}
.fela-b {
  fontWeight: 'bold';
}
```

TxHawks:
If I understand you correctly, Fela has this functionality out of the box, no plugin needed. Just pass a `selectorPrefix` to your renderer:

```js
import { createRenderer } from 'fela';

const renderer = createRenderer({ selectorPrefix: 'fela-', });
```

See https://fela.js.org/docs/advanced/RendererConfiguration.html

houd1ni:
@TxHawks  Thank you! Yes, it's exactly what I thought of.

So, if one has some classnames inference, maybe, he should pick this plugin occasionally instead of solving it by `selectorPrefix`. How about a line in this README warning about it?

TxHawks:
I don't think I understand what the need for a plugin is here. I wouldn't mind a note in the README, though it is pretty clearly documented in the link I sent above.

houd1ni:
@TxHawks Clearly indeed. The case I'm asking, that fela could be easily used without remembering all these options, and one could easily face the problem of class collision and resolve it by the plugin instead of looking for the right solution for him/her. Sometimes it leads to stupid mistakes ;)

TxHawks:
Sure, send a PR.

[...]

PR #689, opened 2019-03-14
Update plugin-isolation's README.md - avoids misuse

houd1ni:
Added a warning line, supposed to avoid misuse and point to the main documentation, that could be partially overlooked while setting up.

Closes https://github.com/rofrischmann/fela/issues/688


rofrischmann:
Thank you!

PR #700, opened 2019-04-10
Sort styles’ keys before rendering class names

HugoGiraudel:
## Description

There is an issue with server-side rendering due to keys being sorted differently on Node and the client (non-Blink based). This pull-request intends to sort object keys before outputting to class names to avoid that React mismatch.

## Packages

- fela@10.2.4

## Versioning

Patch

#### Quality Assurance

  Some tests are not passing but they don’t appear to be related to my changes. 

#### Changes

### DIFF
--- a/packages/fela/src/__tests__/sortObjectKeys-test.js
+++ b/packages/fela/src/__tests__/sortObjectKeys-test.js
@@ -1,0 +1,10 @@
+import sortObjectKeys from '../sortObjectKeys'
+
+describe('Sorting object keys', () => {
+  it('should sort keys in an object', () => {
+    const input = { b: 2, a: 1 }
+
+    expect(JSON.stringify(input)).not.toEqual('{"a":1,"b":2}')
+    expect(JSON.stringify(sortObjectKeys(input))).toEqual('{"a":1,"b":2}')
+  })
+})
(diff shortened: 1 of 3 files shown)


sgrishchenko:
I think this is an incorrect solution to the problem. Quote from [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Deleted_added_or_modified_properties):

> A for...in loop iterates over the properties of an object in an **arbitrary** order

This means that in accordance with the specifications in whatever order we put the properties in the object, there are no guarantees that the next time these keys will be returned in this order. The problem is exacerbated by the server-side rendering context. We already have 2 different platforms: one on the server, the other on the client.

I think the right decision should be a function that returns not a new object with a different order of keys, but an array of keys in the correct order: 
```js
export default function getSortedObjectKeys(input: Object): string[] {
   return Object.keys(input).sort((a, b) => a.localeCompare(b))
}

// ...

let sortedStyleKeys = getSortedObjectKeys(style)
// most performant implementation of cycle as I know
for (let i = 0, length = sortedStyleKeys.length; i < length; i += 1) {
  const property = sortedStyleKeys[i]
  ...
}
```

rofrischmann:
@HugoGiraudel I'll close this one for now? We can discuss it next week in Berlin and probably find a solution to your issues. Maybe #707 is related as well?

PR #703, opened 2019-04-23
feat(renderer): add `targetDocument` option

### LINKED ISSUE #693: Support style injection target in renderer
**Type**: Feature

| Packages | Version |
| -------- | ------- |
| fela     | 6.0.x   |

-------------------
#### Description

I plan on using https://github.com/seek-oss/playroom. It requires to inject the styles into iframes which have the actual relevant dom. But the actual components get executed outside of the iframe and renders the actual from into each iframe, so all styles land in the parent iframe and the child iframes don't inherit them.
styled-components supports this https://github.com/seek-oss/playroom#css-in-js-support so maybe this is also something fela could support.

#### Proposal
The renderer would need an option to tell it the target it should inject the styles.
And use it probably somewhere here: https://github.com/rofrischmann/fela/blob/master/packages/fela-dom/src/dom/connection/createNode.js#L12

layershifter:
Fixes #693.

## Description
This PR adds an `targetDocument` option to `createRenderer()` that allows to render styles properly to child windows: https://github.com/stardust-ui/react/issues/1178

## Packages

- `fela`
- `fela-dom`

## Versioning
Minor

#### Quality Assurance

#### Changes

### DIFF
--- a/packages/fela-dom/src/dom/connection/createNode.js
+++ b/packages/fela-dom/src/dom/connection/createNode.js
@@ -7,11 +7,12 @@ export default function createNode(
   nodes: Object,
   score: number,
   { type, media, support }: NodeAttributes,
+  targetDocument: Document,
   id?: string = ''
 ): Object {
-  const head = document.head || {}
+  const head = targetDocument.head || {}
 
-  const node = document.createElement('style')
+  const node = targetDocument.createElement('style')
   node.setAttribute('data-fela-id', id)
   node.setAttribute('data-fela-type', type)
   node.type = 'text/css'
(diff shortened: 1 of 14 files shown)


packages/fela-dom/src/dom/connection/createNode.js
```diff
@@ -7,11 +7,12 @@ export default function createNode(
   nodes: Object,
   score: number,
   { type, media, support }: NodeAttributes,
+  targetDocument: Document,
```
layershifter: Order of params is changed because `targetDocument` is required

layershifter:
This PR actually adds changes to renderer 👎 Another approach is:
- accept `target` prop on `RendererProvider`
- pass it to `rehydrate()` and `render()` as second param
- pass it to `createSubscription()` and `getNodeCache()`

@rofrischmann can I hear your feedback? 👂

layershifter:
Closing in favor of #704.

PR #704, opened 2019-04-23
feat(RendererProvider): add `target` prop

layershifter:
## Description

Another approach from https://github.com/rofrischmann/fela/pull/703#issuecomment-485756244.

## Example

```jsx
<RendererProvider target={document} />
```

## Packages
- `fela`
- `fela-bindings`
- `fela-dom`
- `react-fela`

## Versioning
Minor

## Checklist

#### Changes

### DIFF
--- a/docs/advanced/RendererConfiguration.md
+++ b/docs/advanced/RendererConfiguration.md
@@ -13,6 +13,7 @@ We might introduce more configuration options with future releases, so be sure t
 | enhancers  | *(Array?)* |  |  A list of [enhancers](../advanced/Enhancers.md) to enhance the renderer
 | mediaQueryOrder | *(Array?)* |  | An explicit order in which `@media` queries are rendered |
 | rendererId | *(string?)* |  | An optional unique identifier that will prefix your animation names and will be added as the value of a `data-fela-id` attribute on `<style>` nodes. An example use case: the recommended way is to have only one renderer, but there are some specific cases when you will need to have multiple renderers. This option allows you avoid conflicts in generated CSS rules via prefixing animation names and separated `<style>` tags. |
+| targetDocument | *(Document?)* | document | Allows to pass you a custom document to insert style rules, useful when you're are rendering to child windows. 
 | supportQueryOrder | *(Array?)* |  | An explicit order in which `@supports` queries are rendered |
 | selectorPrefix | *(string?)* |  | Prepend a static prefix to every generated class and keyframe. It must only consist of `a-zA-Z0-9-_` and start with `a-zA-Z_`. |
 | filterClassName | *(Function?)* | `cls => cls.indexOf('ad') !== -1` | Filter-function to filter used class names |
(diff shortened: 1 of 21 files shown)


layershifter:
@rofrischmann If you will agree with these changes, I will update docs and will add more UTs 🙄

layershifter:
Other opened question: how we will handle unsubscribe to avoid leaking subscriptions ❓

rofrischmann:
Can you go into detail on how the new stuff works? I have barely time to test it out myself and would love to have a short description of how the target stuff works, what changed and what the default is.

layershifter:
Sure.

### L1 - `fela-dom` & `documentId`

I added the `documentId` that it's stored on `document.head` as attribute, it allows:
- handle it in references as `string` to get nodes from cache (`getNodeFromCache.js`)
- access it in `render()` (`render.js`) to ensure that new subscription is needed

Actually, we need it only to properly access correct nodes from existing cache (`renderer.nodes`).

### L2 - `fela-dom` & `target`

I added the `target` to `render()` and `rehydrate()` functions. It's an optional param that allows you to specify the `document` to render styles.

If `target` will be passed in `render()` we will ensure that we have a subscription for this document. If not, we will create a new subscription and will restore everything from cache with styles (`renderer.cache`).

**ISSUE 1**: there are no way to handle multiple subscriptions because `renderer.updateSubscription` is a function.
**SUGGESTION:** In a separate PR make it an object and rename to `updateSubscriptions`. It's not exposed as public API, so it is not breaking change. With this change we will be able to store subscriptions by `documentId`:
```
renderer.updateSubscriptions = {
  '0': updateSubscription(),
  '1': updateSubscription(),
}
``` 

**ISSUE 2**: As I mentioned before https://github.com/rofrischmann/fela/pull/704#issuecomment-486571394, we need somehow to clean subscriptions and nodes from cache. We can implement this in bindings via `componentWillUnmount()`

### L3 - `fela-bindings` & `target`

In `RendererProviderFactory` we simply pass `target` to `render()` and `rehydrate()` from `fela-dom`, so we don't have any changes in `renderer`.

rofrischmann:
Alright, I see. So according to the issues: Is this something we need to fix now or do you think we should have that in a separate PR? Do we really need multiple subscriptions?

layershifter:
>  Is this something we need to fix now or do you think we should have that in a separate PR?

We definitely need to fix this before merging...

> Do we really need multiple subscriptions?

For me it looks like the best option because each created subscription is atomic. If we will remove a document it means that we should simply remove subscription, i.e. unmount `RendererProvider` with defined target means that matching document was removed and subscription can be removed.

Another approach is to rewrite `createSubscription` to handle cached documents: we will have a loop with documents inside of created subscription and inside of it we can ensure that document still exists.

```jsx
renderer.cache.documents.forEach(document => {
   if (!documentIsPresent(document)) // remove document from cache

   const node = getNodeFromCache(change, renderer, document)

   switch (change.type) {}
})

rofrischmann:
Hmm, so for the issues: Feel free to send a PR for that! (or update this one)
But I'm still not sure on the multi subscriptions. I think I don't want those - if you have multiple subscriptions you're pretty sure doing something wrong anyways.

Issue #713, opened 2019-06-12
Prefix Keyframe name


josemussa:
**Type**: Feature

| Packages | Version |
| -------- | ------- |
| fela     | latest   |

-------------------
#### Description
Although not recommended we have situations were multiple renderer instances are needed, a script tag like a chat box or any other secondary application on the same page.

So far we have been able to circumvent all clashes by adding a custom `selectorPrefix` but when it comes to animations we still have problems with keyframes, my suggestion for this is to add something similar to `selectorPrefix` to the [generateAnimationName](https://github.com/rofrischmann/fela/blob/master/packages/fela/src/generateAnimationName.js) function.

For proper backward compatibility, we need to use something different to `selectorPrefix`, I see that Fela is already using the name `keyframePrefixes` for other functionality so `keyframePrefix` might be a bit confusing, maybe `animationsPrefix`?

**Suggested implementation**
```js
/* @flow */

export default function generateAnimationName(id: number, prefix?: string = ''): string {
  return prefix + 'k' + id
}
```

PR #716, opened 2019-06-12
fix: adds selectorPrefix to animationName

josemussa:
https://github.com/rofrischmann/fela/issues/713
## Description
According to the documentation, `selectorPrefix` should work with keyframes.
> Prepend a static prefix to every generated class and **keyframe**. It must only consist of a-zA-Z0-9-_ and start with a-zA-Z_.

I made a more complex solution [here](https://github.com/rofrischmann/fela/pull/715) that adds `animationPrefix` as a configuration key but we can totally reuse the `selectorPrefix` as the docs originally state

## Packages
fela

- 

## Versioning
Patch

## Checklist

#### Changes


weserio:
I think we'll stick with this version. Don't want to use another prefix here tbh.

PR #718, opened 2019-06-18
feat(RendererProvider): add `target` prop

layershifter:
## Description

Another approach (from #704) for fixing #693.

### `RendererProvider`

Now this component accepts `target` prop like in `styles-components` [StyleSheetManager.js#L25](https://github.com/styled-components/styled-components/blob/v4.3.1/packages/styled-components/src/models/StyleSheetManager.js#L25), it allows to specify a `Document` where style nodes will be attached.

### `subscribeDocument()` & `documentRefs`

This method is added to handle cases with passed documents, this implementation allows us to have a single subscription (difference from #704). `subscribeDocument()` returns and object with `unsubscribe()` like `subscribe()` does.

```js
const r = createRenderer()

r.subscribeDocument(childDocument)
felaDOM.render(r)
// styles will be renderer to main `document` and `childDocument`
```

However, we need to store `Document` object somewhere, I propose:

```js
{
  target: Document
  refId: string
  refCount: number
}
```

We need to have `refCount` to know when we should remove matching `Document` and clear `renderer.nodes` to avoid memory leaks. `refId` is used by build `reference` key in `renderer.nodes` because we can't rely on indexes.

## TODO

- [ ] update docs
- [ ] add unit tests

## Example

```jsx
<RendererProvider target={childDocument} />
```

## Packages

- `fela-bindings`
- `fela-dom`
- `fela`
- `react-fela`

## Versioning
Minor

#### Quality Assurance

#### Changes

### DIFF
--- a/packages/fela/src/getDocumentRefIndex.js
+++ b/packages/fela/src/getDocumentRefIndex.js
@@ -1,0 +1,17 @@
+import arrayReduce from 'fast-loops/lib/arrayReduce'
+
+import type { DOMRendererDocumentRef } from '../../../flowtypes/DOMRenderer'
+
+const getDocumentRefIndex = (
+  documentRefs: DOMRendererDocumentRef[],
+  target: Object
+) =>
+  arrayReduce(documentRefs, (accIndex, documentRef, index) => {
+    if (documentRef.target === target) {
+      return index
+    }
+
+    return accIndex
+  })
+
+export default getDocumentRefIndex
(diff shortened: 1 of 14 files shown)


packages/fela/src/getDocumentRefIndex.js
```diff
@@ -0,0 +1,17 @@
+import arrayReduce from 'fast-loops/lib/arrayReduce'
+
+import type { DOMRendererDocumentRef } from '../../../flowtypes/DOMRenderer'
+
+const getDocumentRefIndex = (
```
layershifter: Can be replaced with `.findIndex()` but there is no such function in `fast-loops` and it's not supported by IE11

PR #733, opened 2019-09-04
Fix clear on removed nodes

per2plex:
## Description
I'm using Fela in conjunction with Storybook and I'm having the problem that Storybook sometimes removes the Fela style nodes on hot reload before I can call `clear` which would end up in an error, this PR solves this problem by checking if the `parentNode` exists before trying to remove the node. 

This is not directly related to Storybook and I can see other cases where this could happen.

**EDIT: Found out that the renderer got recreated in the Storybook config file on hot reload. Moving the renderer into its own module solved the problem without needing this fix, this fix might still be applicable in other cases tho, if you don't think this is something Fela should handle you can close this PR**

## Example
N/A

## Packages

- fela-dom

## Versioning
None

#### Quality Assurance

#### Changes


robinweser:
I think we don't want that fix being part of fela itself, since it's more like an hot reload specific issue.
But I'll definitely add some info to the docs for hot reloading! Thanks for reporting :)

PR #734, opened 2019-09-13
Respect _className in combineRules

robinweser:
## Description
This PR adds the ability to combine _className keys in `combineRules`.

## Example

```javascript
const foo = () => ({
  _className: "foo",
  color: "blue"
})

const bar = () => ({
  _className: "bar",
  fontSize: 15
})

const combined = combineRules(foo, bar)()

combined === {
  _className: "foo bar",
  color: "red",
  fontSize: 15
}
```

## Versioning
Major

Even though minimal, this actually is a major change.

## Checklist

#### Changes

### DIFF
--- a/packages/fela/src/__tests__/combineRules-test.js
+++ b/packages/fela/src/__tests__/combineRules-test.js
@@ -72,7 +72,7 @@ describe('Combining rules', () => {
   })
 
   it.only('should merge _className as a special key', () => {
-    const rule = props => ({
+    const rule = () => ({
       _className: 'foo bar',
       color: 'red',
       padding: 10,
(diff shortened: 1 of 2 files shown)


TxHawks:
LGTM 🚀

packages/fela/src/__tests__/combineRules-test.js
```diff
@@ -70,4 +70,31 @@ describe('Combining rules', () => {
       padding: 20,
     })
   })
+
+  it.only('should merge _className as a special key', () => {
```
layershifter:
```suggestion
  it('should merge _className as a special key', () => {
```

oops

robinweser: Ops indeed always doing that same mistake again and again

PR #756, opened 2019-12-12
Allow enhancers to control identifier generation

10xjs:
## Description

Move the existing class and animation name creation logic to separate methods on the renderer object which can be overridden by enhancers.

This enables the creation of enhancers which modify the identifier name generation without needing to re-implement the full `_renderStyleToClassNames` or `renderKeyframe` methods.

Importantly, this helps unblock concerns like https://github.com/robinweser/fela/issues/493 by shifting more control to the consumer.

## Example

Here's a basic example of how an enhancer could leverage this change:

```javascript
function generateAtomicClassName(property, value, pseudo, media, support) {
  // Return a unique deterministic class name derived from the current style.
};

function atomicClassNames() {
  return (renderer) => {
    renderer.generateClassName = (property, value, pseudo, media, support) => {
      return generateAtomicClassName(property, value, pseudo, media, support);
    };

    return renderer;
  };
};

  enhancers: [atomicClassNames()],
});
```

## Packages
`/packages/fela`

## Versioning
Patch

## Checklist

#### Changes

### DIFF
--- a/packages/fela/src/createRenderer.js
+++ b/packages/fela/src/createRenderer.js
@@ -89,6 +89,10 @@ export default function createRenderer(
       return ++renderer.uniqueRuleIdentifier
     },
 
+    getNextKeyframeIdentifier() {
+      return ++renderer.uniqueKeyframeIdentifier
+    },
+
     renderRule(rule: Function, props: Object = {}): string {
       return renderer._renderStyle(rule(props, renderer), props)
     },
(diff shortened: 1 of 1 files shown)


10xjs:
@robinweser Have you had a chance to look at this?

robinweser:
So I'm just reviewing this. Looks good to me, but I'm not sure if you'll be able to implement deterministic class names that way as that's not really possible without super long class names and/or hash collision that might cause server/client differences.

PR #763, opened 2020-02-06
Add umd builds for react-fela

IgorBabkin:
```javascript
cd packages/react-fela && yarn build:umd
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ node_modules/
 npm-debug.log
 lib/
 es/
+umd/
 coverage/
 _book
 
--- a/package.json
+++ b/package.json
@@ -75,6 +75,11 @@
     "lerna": "^3.4.3",
     "prettier": "^1.18.2",
     "recompose": "^0.25.0",
-    "rimraf": "^2.6.1"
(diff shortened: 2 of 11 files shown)


robinweser:
Would you be open to introduce that change for every package? Should be easy to do that with Lerna. Externals should be `React`, `Preact` and `Inferno` as of today.

Would love to have that =) 
Thanks for your PR anyways!

IgorBabkin:
I've introduced common solution how to simply add UMD bundling for any package. And also have added bundling for packages I need in my projects. I think we need to add UMD on demand. Some of package no need UMD because they are lightweight like `fela-plugin-fallback` or `fela-unit`.

IgorBabkin:
Let me know for what packages you want to have UMD?

IgorBabkin:
@robinweser Should I add anything?

robinweser:
Hey Igor,
It looks fine for now. I might refactor it later to an external script just like the build script with a single reused webpack config if possible. So we don't have to add it to each package, but rather build from the root, but I'll merge it now to release the changes :)

PR #781, opened 2020-04-16
docs(fallback): improve example

CanRau:
I think the rules in the readme of `fela-plugin-fallback-value` should be swapped to show a real-world example and have `#ccc` be fallback if `rgba` isn't supported in the browser.
Please correct me if I'm wrong, or misunderstand the purpose.

For reference: https://modernweb.com/using-css-fallback-properties-for-better-cross-browser-compatibility/

### DIFF
--- a/packages/fela-plugin-fallback-value/README.md
+++ b/packages/fela-plugin-fallback-value/README.md
@@ -30,20 +30,20 @@ const renderer = createRenderer({
 #### Input
 ```javascript
 {
-  color: [ 'rgba(0, 0, 0, 0.5)', '#ccc']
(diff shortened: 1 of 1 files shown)


CanRau:
Maybe I'm wrong, couldn't verify, I just guessed that the example in `#### Output` should be 
```javascript
{
  color: '#ccc;rgba(0, 0, 0, 0.5);'
}
```
instead of 
```javascript
{
  color: 'color:#ccc;rgba(0, 0, 0, 0.5);'
}
```
removing the duplicated(?) `color:`

robinweser:
You're actually right, thanks for finding that!

CanRau:
You're welcome 🤗

PR #786, opened 2020-04-19
Add missing `return` to README for React example

jjenzz:
Does what it says on the tin. I spotted this whilst trying to get Fela working on Codesandbox.


robinweser:
Whoops, thanks again! must have copied that from reason^^

PR #803, opened 2020-08-10
Performance improvements on rule rendering

robinweser:
## Description
This PR removes `camelCaseProperty` on each render which can boost perf by up to 20%.
It's still a breaking change as it now officially disallows dash-cased property names.
We might want to provide a plugin that transforms those beforehand, so people have a smooth migration path.

## ToDo
- [x] Make sure rehydration works properly
 
## Packages

- fela
- fela-utils

## Versioning
Major

## Checklist

#### Changes


TxHawks:
What would happen when needing to manually add prefixed props, such `-webkit-tap-highlight-color`?
Currently the only way to add such properties is to add them as a qouted kebab-case property.

robinweser:
Hmm, no. You can just do `WebkitTabHighlightColor` as well.
There's really no need for kebab-cased properties at all.

PR #773, opened 2020-03-30
Add fela-dom UMD bundle

jp7837:
## Description
Add fela-dom UMD bundle, similar to #763

## Packages
List all packages that have been changed.

- fela-dom

## Versioning
None

## Checklist

- [ ] The code was formatted using Prettier (`yarn run format`)
- [ ] The code has no linting errors (`yarn run lint`)
- [ ] All tests are passing (`yarn run test`) 
- [ ] There are no flow-type errors (`yarn run flow`)

#### Changes
If one of the following checks doesn't make sense due to the type of PR, just check them.

- [ ] Tests have been added/updated
- [ ] Documentation has been added/updated
- [ ] My changes have proper flow-types

PR #805, opened 2020-08-20
[fela-dom] Add UMD bundle to NPM package

jp7837:
[fela-dom] Add UMD bundle to NPM package

Missed change from #773


jp7837:
@robinweser @IgorBabkin

PR #814, opened 2020-09-25
feat: add `fela-plugin-fullscreen-prefixer`

chmelevskij:
## Description
Add plugin for prefixing `:fullscreen` property. Pretty much copy of the `fela-plugin-placeholder-prefixer`

## Packages

- `fela-plugin-fullscreen-prefixer`

## Versioning
None? or Major?

## Checklist

#### Changes

### DIFF
--- a/packages/fela-plugin-fullscreen-prefixer/LICENSE
+++ b/packages/fela-plugin-fullscreen-prefixer/LICENSE
@@ -1,0 +1,21 @@
+The MIT License (MIT)
+
+Copyright (c) 2017-2020 Robin Weser
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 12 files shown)


scottckr (commented):
You got 7 occurences of `placeholder` remaining from copying this plugin :)

packages/fela-plugin-fullscreen-prefixer/src/__tests__/placeholderPrefixer-test.js
```diff
@@ -0,0 +1,31 @@
+import fullscreenPrefixer from '../index'
```
scottckr, line 1: This file is still called `placeholderPrefixer-test`

alizeait:
How about a plugin that takes a pseudo element and an array of prefixes instead of duplicating the same package for every pseudo element? Could be something like `fela-plugin-pseudo-prefixer`:
```javascript
export default function pseudoPrefixer(pseudoElement,prefixes) {
  return customProperty({
    [pseudoElement]: value =>
      arrayReduce(
        prefixes,
        (style, prefix) => {
          style[prefix] = value
          return style
        },
        {}
      ),
  })
}
```
Then do:
```javascript
  plugins: [ pseudoPrefixer(':fullscreen',[
  ':-webkit-full-screen',
  ':-moz-full-screen',
  ':-ms-fullscreen',
  ':full-screen',
  ':fullscreen',
])
]})
```
We can also use this plugin to create the `fela-plugin-placeholder-prefixer` plugin. This way we keep a single set of tests in one place and avoid duplicating code. Would also make it easier to create more prefixes in the future for other elements.

chmelevskij:
Yeah, probably makes more sense to have more generic one.

But there is value in having `fela-plugin-fullscreen-prefixer` or other property ones. I'm not really keen on going and fishing out prefixes every time I need to prefix something.

alizeait:
Definitely keep `fela-plugin-fullscreen-prefixer` and `fela-plugin-placeholder-prefixer`, as they would also use the more generic plugin to provide those prefixes. Something like:
```javascript
export default function fullscreenPrefixer(){
  return pseudoPrefixer(':fullscreen',[
  ':-webkit-full-screen',
  ':-moz-full-screen',
  ':-ms-fullscreen',
  ':full-screen',
  ':fullscreen',
])
}
```

PR #827, opened 2020-12-06
propertyPriority config + fela-enforce-longhands

robinweser:
## Description
This PR implements a new configuration option `propertyPriority`.
It is a map of property-priority pairs and determines which properties should have a higher specificity score.
The default priority for every rule is `1`, so in order to increase that, values > 1 need to be passed.
This can be used to implement certain shorthand-longhand priorities.

This PR also adds a new enhancer `fela-enforce-longhands` which implements the `propertyPriority` configuration in a way that longhands always overwrite their counterpart in shorthand property.

## ToDo
- [x] add all longhands to fela-enforce-longhands
- [ ] add option to choose whether directional or "level 2" longhands are prioritized for borders (e.g. `border-left` vs. `border-width`)
- [x] add documentation once #818 is done
  - [x] How it works
  - [x] Link to CSS3 Selector Spec
- [x] add proper tests

## Example

```javascript
import { createRenderer } from "fela" 

  propertyPriority: {
    paddingLeft: 2,
    paddingRight: 2,
    paddingBottom: 2,
    paddingTop: 2
  }
})

// => a b
const className = renderer.renderRule(() => ({
  paddingLeft: 10,
  padding: 20
})
```

```css
.a.a {
  padding-left: 10px
}

.b {
  padding: 20px
}
```
> Even though `padding: 20px` is renderer **after** `padding-left:10px` it will still be overwritten due to CSS specificity.

## Packages

- fela
- fela-dom
- fela-utils

## Versioning
Minor

## Checklist

#### Changes


packages/fela-dom/src/dom/rehydration/rehydrateRules.js
```diff
@@ -9,6 +9,21 @@ function escapeRegExp(string) {
...
 }
 
+// we reuse regexes for improved performance
+const regexMap = {}
+function getRegex(specificityPrefix) {
+  if (!regexMap[specificityPrefix]) {
+    regexMap[specificityPrefix] = new RegExp(
+      `${escapeRegExp(
+        specificityPrefix
+      )}[.]([0-9a-z_-]+)([.][0-9a-z_-]+){0,}([^{]+)?{([^:]+):([^}]+)}`,
```
HugoGiraudel, line 19: I think you could slightly simplify it like this:
```suggestion
      )}[.]([\w-]+)([.][\w-]+){0,}([^{]+)?{([^:]+):([^}]+)}`,
```

packages/fela-utils/src/generateCSSSelector.js
```diff
@@ -2,7 +2,10 @@
 export default function generateCSSSelector(
   className: string,
   pseudo: string = '',
-  specificityPrefix?: string = ''
+  specificityPrefix?: string = '',
+  propertyPriority?: number = 1
 ): string {
-  return `${specificityPrefix}.${className}${pseudo}`
+  const classNameSelector = `.${className}`.repeat(propertyPriority)
```
HugoGiraudel, line 8: I don’t know how safe this is is going to be if I’m being honest. I was hoping we could solve that at the resolver level rather than within the stylesheet.

robinweser: Safe in what regard?

robinweser:
Merging this with a follow up #830

PR #833, opened 2020-12-14
Add a recipe for using dev utils as devDependencies

HugoGiraudel:


robinweser:
Nice addition! Thanks for submitting that :)

PR #891, opened 2022-02-22
fix(fela-enforce-longhands): fix typo in property name

layershifter:
## Description
It fixes typo in CSS property name.


robinweser:
Oh.. Luckily, backgroundRepeat is not used that often 😅 
Thanks for the quick fix!

PR #894, opened 2022-02-28
Theme-aware static rendering

robinweser:
## Description

Added a third argument `props` to `renderer.renderStatic` which are used when processing plugins.
This is mostly used when processing styles with plugins like theme-value or responsive-value where we want the `theme` being present in our props.

Additionally, I added a `staticStyle` function to `useFela` which passes the props including theme by default. This works exactly the same as the props passing in `css`.

## Example

```javascript
const theme = {
  colors: {
    primary: 'blue',
  },
}

// assuming fela-plugin-theme-value is used with just backgroundColor: theme => theme
renderer.renderStatic(
  {
    backgroundColor: 'colors.primary',
  },
  'body',
  { theme }
)
```

## Packages

- fela
- fela-bindings
- react-fela
- preact-fela

## Versioning

Minor

#### Quality Assurance

#### Changes


chrisrzhou:
@robinweser, this looks great and I think will allow plugins needing additional props to work.

I think my original issue wasn't due to props-based plugins (e.g. `theme-value`) not working, but rather some behaviors of certain plugins not working (e.g. `responsive-value`).  In https://github.com/robinweser/fela/issues/876#issuecomment-813865881, I think the main culprits that were not working were:
- Nested queries
- Pseudo selectors

The above are usually supported in inlined style definitions when using Fela, but when using with `renderer.renderStatic` some functionality was dropped.  Not sure if you see the same issue on your end but see if it's worth confirming with a simple example setting up nested queries and pseudo selectors.

Thanks,
Chris

robinweser:
@chrisrzhou Yeah, this does not introduce nested behaviour to static rendering yet. That's a feature I want to provide with an enhancer in the future, but didn't have time yet to do.

PR #897, opened 2022-04-04
update bindings dependency for compare to move off version with node-fetch vulnerability

micahgodbolt:
react-addons-shallow-compare 15.6.2 has a dependency on an old version of node-fetch with a known vulnerability


robinweser (approved):
Thanks for that fix!

PR #899, opened 2022-04-20
Update link to the Patterns talk

pepelsbey:
## Description

The “CSS in JS: Patterns“ talk is not available with the old link, so here’s the new one.

- Old: https://www.webexpo.net/prague2018/talk?id=css-in-js-patterns
- New: https://www.webexpo.net/prague2018/talk/css-in-js-patterns


robinweser:
Thanks again, @pepelsbey!

PR #901, opened 2022-04-25
Update rules docs: pseudo elements, feature queries

pepelsbey:
## Description

I’ve been reading the docs and discovered a few places that could be improved. This PR is focused on basics/rules.mdx, but I have a few more improvements in other parts. Please let me know if I should use other PRs for them or combine everything in one place.

Here’s the following:

- `::before` is a pseudo element, not pseudo class
- It’s preferable to use double colons to differentiate elements from classes (according to spec)
- It’s worth mentioning that pseudo classes/elements could be combined in a single string, not just nested
- `@supports` is a part of Feature queries, not Support queries

#### Quality Assurance

#### Changes

### DIFF
--- a/website/docs/latest/basics/rules.mdx
+++ b/website/docs/latest/basics/rules.mdx
@@ -56,13 +56,17 @@ const rule = (props) => ({
     color: 'black',
   },
   // make sure you are using nested quotes to set the content.
-  ':before': {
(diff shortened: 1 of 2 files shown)


robinweser (approved):
Nice improvements, thank you!

PR #918, opened 2024-03-18
fix: include children prop in ruleProps

1000hz:
## Description

This PR allows style rules to be conditionally applied based on (for example) the presence or absence of any children passed to the component. Previously, `props.children` was not available within a style rule function.

I wasn't sure where the best place to include a test for this is. There don't seem to be any relevant tests within `fela-bindings`, but let me know if you'd like me to include a test in one of the downstream packages and I'll be happy to add one.

## Packages

`fela-bindings`

## Versioning

Patch

#### Quality Assurance

#### Changes

### DIFF
--- a/packages/fela-bindings/src/createComponentFactory.js
+++ b/packages/fela-bindings/src/createComponentFactory.js
@@ -77,6 +77,7 @@ export default function createComponentFactory(
             theme: _felaTheme,
             as,
             id,
+            children,
           }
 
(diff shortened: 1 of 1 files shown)


robinweser:
@1000hz Thanks for the PR! Great addition actually, weird that no one else requested that before.
And sorry for the delay, I‘ve been on vacation for most of the past month.
Will do a release first thing tomorrow morning!

PR #923, opened 2025-01-25
feat: Allow ampersand suffixes

### LINKED ISSUE #702: Can't style based on global parent class
**Type**: Bug

| Packages | Version |
| -------- | ------- |
| fela     | 6.1.x & 10.2.x (tested both)  |
| react-fela | 7.3.x & 10.2.x |

-------------------
#### Description

I'm trying to integrate with a third-party library ([Tether](http://tether.io/)), which adds dynamic classes to a container component. These classes can be added and removed at any time, based on its position adapting to window size.

I'd like to style a component (entirely within the Tether container) based on the container's current classes. This doesn't seem to be possible with Fela.

[The docs say](https://fela.js.org/docs/basics/Rules.html):
> While we only support the above selectors for a reason, we also understand that there might be some edge cases (mostly with third-party libraries) where you want to use other selectors as well. Therefore we provide the &-prefix for nested selectors.

From what I can tell, `&` only works when it's at the start of the selector (hence calling it "&-**prefix**"). But I need to use it _after_ the Tether class, like `.tether-element-attached-top &`. This works in tools such as Sass and Styled Components.

This is not just a problem when using Tether, but affects integration with any library that adds CSS classes to the `body` element as part of its API (e.g. font preloading scripts which add a `fonts-loaded` class to `body` for conditional styling).

As a simplified example, I have a component with these basic styles:

```js
{
  background: 'red',
  '.tether-element-attached-top &': {
    background: 'green',
  },
  '.tether-element-attached-bottom &': {
    background: 'blue',
  },
}
```

On its own the component should have a red background. If used within a Tether element, it should be either blue or green, depending on Tether's positioning calculations. When used within Tether, the markup is effectively this:

```html
<div class="tether-enabled tether-element-attached-right tether-target-attached-right tether-element-attached-bottom tether-target-attached-top">
  <div class="a">This is my component element</div>
</div>
```

However when I do this, the element is always red, and Fela adds the following warning in the console:
`The object key ".tether-element-attached-top &" is not a valid nested key in Fela.`

stephenh:
## Description

Allows nested rules like:

```js
".Group:hover &": { color: "white" },
```

Where the `&` comes at the end.

Output:

```css
.Group:hover .a { color: "white" }
```

This effectively fixes #702 , and mimics the Tailwinds capability of styling a child based on "some parent" being hovered. 

I definitely understand the "nested selectors are terrible", and we have ~very few in our codebase :sweat_smile: (that currently uses Emotion), but this "change my style based on my parent", particularly with a CSS-driven `:hover` (instead of tracking parent hover state in JS), seems like a valid usage. 

## Packages

* fela
* fela-utils

## Versioning

Minor

#### Quality Assurance

#### Changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -77,5 +77,6 @@
         ]
       }
     }
-  }
+  },
+  "packageManager": "pnpm@8.7.6+sha1.a428b12202bc4f23b17e6dffe730734dae5728e2"
 }
--- a/packages/fela-integration/src/fela_fela-tools/__tests__/__snapshots__/integration-test.js.snap
+++ b/packages/fela-integration/src/fela_fela-tools/__tests__/__snapshots__/integration-test.js.snap
@@ -16,6 +16,8 @@ exports[`Fela with Fela Tools integration Rendering rules should remove undefine
 
 exports[`Fela with Fela Tools integration Rendering rules should render any nested selector with the &-prefix 1`] = `".a{color:red}.b~#foo{color:blue}.c .bar{color:green}"`;
 
+exports[`Fela with Fela Tools integration Rendering rules should render any nested selector with the &-suffix 1`] = `".a{color:red}.bar:hover .b{color:green}"`;
+
 exports[`Fela with Fela Tools integration Rendering rules should render attribute selectors 1`] = `".a{color:red}.b[bool=true]{color:blue}"`;
 
 exports[`Fela with Fela Tools integration Rendering rules should render attribute selectors and add specificity prefix 1`] = `".parent .a{color:red}.parent .b[bool=true]{color:blue}"`;
--- a/packages/fela-integration/src/fela_fela-tools/__tests__/integration-test.js
+++ b/packages/fela-integration/src/fela_fela-tools/__tests__/integration-test.js
@@ -142,7 +142,6 @@ describe('Fela with Fela Tools integration', () => {
       expect(renderToString(renderer)).toMatchSnapshot()
     })
 
-
     it('should render media queries', () => {
       const rule = () => ({
         color: 'red',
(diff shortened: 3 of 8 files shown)


stephenh:
Huh; I thought this was a pretty straight-forward approach, but I'm getting test failures in rehydration:

image

That shows the `&` getting dropped; but if anything I thought it would "show up more" because I removed the `normalizeNestedProperty(property)` that used to be dropping the `&`, but now is leaving it in place, so it can later be swapped by the `generateCSSSelector` logic.

stephenh:
I reverted to a simpler approach, that doesn't use `replace` and instead uses a `pseudoPrefix` flag to mean "put the class name at the end of `pseudo`".

This passes the current test suite, b/c `&` is eagerly removed from the pseudo selector, and so not in the cache keys, and so the hydration test suite passes.

That said, the `&` at the end will still break rehydration b/c right now rehydration is able to look at a CSS rule, and use the `DECL_REGEX` to parse back out what the original specificity prefix, class name, psuedo, etc., are all merely by looking at the rule.

However the regexes currently assumes "the class name is at the beginning" (technically after the specificity prefix) and I'm not sure it's possible for the regex to handle "actually the class name may be at the beginning or end" / "before or after the pseudo selector", i.e.

* `& #id > .foo ~ bar` (current rehydration test) would be `.a #id > .foo ~ bar`, but
* `#id > .foo ~ bar &` (new rehydration test with trailing `&`) would be `#id > .foo ~ bar .a`

This could probably work by adding a `data-fela-amp` type hint to the `node`s, so that rehydration could look it up around here:

```js
    if (rehydrationIndex !== -1) {
      const type = node.getAttribute('data-fela-type') || ''
      const media = node.getAttribute('media') || ''
      const support = node.getAttribute('data-fela-support') || ''
      const css = node.textContent
```

I.e. a `node.getAttribute("data-fela-amp")` that returns `"start"` or `"end"` or `null`, and if it was start/end, the regex could be adjusted appropriately.

This doesn't seem very hard, and is precedence for "rehydration hints" in the `data-fela-support` attribute, so maybe that's the right way to go?

Dunno, personally I wouldn't need rehydration, so maybe getting "ampersand suffixes" supported in rehydration could be left as a todo?

